PR #5, opened 2018-04-16
Update README

stweil:
There is no tesseract-ocr-deu-frak for Tesseract 4.0.
It was replaced by tesseract-ocr-frk.

There is still a tesseract-ocr-deu-frak for Tesseract 3.04,
but it uses a different installation path and does not support LSTM.

In addition to the language models, there are now also script models.

Fixing headers is no longer needed with latest Debian packages and should
also be fixed for Ubuntu soon.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,7 +17,7 @@ Installation
 Required ubuntu packages:
 
 * Tesseract headers (``libtesseract-dev``)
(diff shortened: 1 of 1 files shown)


kba:
Thanks! `deu-frak` was still a remnant from installation against 3.04 headers, the header issue went away with a clean install of 4.00 and the latest dev tesserocr bindings.

PR #8, opened 2018-04-25
adapt to changes in core

kba:
* adapted to changes in core
* word segmenter (within lines only atm)

### DIFF
--- a/ocrd-tool.json
+++ b/ocrd-tool.json
@@ -18,7 +18,20 @@
       "tags": ["Text recognition and optimization"],
       "description": "Recognize text in lines with tesseract",
       "executable": "ocrd-tesserocr-recognize",
-      "step": "recognition/text-recognition"
+      "step": "recognition/text-recognition",
+      "parameter": {
+        "textequiv_level": {
+          "type": "string",
+          "enum": [
+            "page",
+            "region",
+            "line",
+            "word",
+            "glyph"
+          ],
+          "default": "line"
+        }
+      }
     }
   ]
 }
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -1,8 +1,9 @@
 from __future__ import absolute_import
 
 import tesserocr
-from ocrd.utils import getLogger, mets_file_id
-from ocrd import Processor, OcrdPage, MIMETYPE_PAGE
+from ocrd.utils import getLogger, mets_file_id, xywh_from_coordinate_string
+from ocrd.model.ocrd_page import from_file, to_xml, TextEquivType
+from ocrd import Processor, MIMETYPE_PAGE
 from ocrd_tesserocr.config import TESSDATA_PREFIX
 
 log = getLogger('processor.TesserocrRecognize')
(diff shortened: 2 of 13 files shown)


ocrd-tool.json
```diff
@@ -18,7 +18,20 @@
       "tags": ["Text recognition and optimization"],
       "description": "Recognize text in lines with tesseract",
```
wrznr: Is it lines only?

ocrd-tool.json
```diff
@@ -18,7 +18,20 @@
...
       "executable": "ocrd-tesserocr-recognize",
-      "step": "recognition/text-recognition"
+      "step": "recognition/text-recognition",
+      "parameter": {
+        "textequiv_level": {
```
wrznr: I am not sure here: Shouldn't we propagate the `textequiv` contents through all the levels?

ocrd_tesserocr/recognize.py
```diff
@@ -1,41 +1,49 @@
...
-                        # xTODO use binarized / gray
-                        image = self.workspace.resolve_image_as_pil(image_url, line.coords)
-                        tessapi.SetImage(image)
-                        line.textequiv = tessapi.GetUTF8Text()
+                pcgts = from_file(self.workspace.download_file(input_file))
+                # TODO use binarized / gray
+                pil_image = self.workspace.resolve_image_as_pil(pcgts.get_Page().imageFilename)
+                tessapi.SetImage(pil_image)
+                # TODO slow
+                #  tessapi.SetPageSegMode(PSM.SINGLE_LINE)
```
wrznr: We definitely have to check what's wrong here.

ocrd_tesserocr/segment_line.py
```diff
@@ -12,19 +18,27 @@ def process(self):
...
-                page = OcrdPage.from_file(self.workspace.download_file(input_file))
-                image_url = page.imageFileName
-                for region in page.list_textregions():
-                    log.debug("Detecting lines in %s with tesseract", region)
-                    image = self.workspace.resolve_image_as_pil(image_url, region.coords)
+                pcgts = from_file(self.workspace.download_file(input_file))
+                image_url = pcgts.get_Page().imageFilename
+                for region in pcgts.get_Page().get_TextRegion():
+                    log.debug("Detecting lines in %s with tesseract", region.id)
+                    image = self.workspace.resolve_image_as_pil(image_url, polygon_from_points(region.get_Coords().points))
```
wrznr: Wasn't this function renamed?

ocrd-tool.json
kba: At the moment it is. Walk the regions, walk the lines within the regions, recognize text in the bounding box of the line. To make that configurable was the idea of the textequiv_level parameter.

ocrd-tool.json
kba: Yes, `TextEquiv` should be propagated where possible. But you might want to run the recognition on sub-regions, on different levels. If the input PAGE is already zoned on line level, you can (as you implemented) set the PSM to `SINGLE_LINE`. Or you might have an OCR result on the word/glyph level as input and want to recognize only the words to compare results.

Maybe `recognition-level` is less confusing?

ocrd_tesserocr/segment_line.py
kba: The functions have the syntax `X_from_Y`, where X/Y can be
  * `points` a string encoding a polygon: `"0,0 100,0 100,100, 0,100"`
  * `polygon` an array of x-y-tuples of a polygon: `[[0,0], [100,0], [100,100], [0,100]]`
  * `xywh` a dict with keys for x, y, width and height: `{'x': 0, 'y': 0, 'w': 100, 'h': 100}`

`points` is for PAGE

`polygon` is what opencv2 expects

`xywh` is what tesserocr expects/produces.

PR #14, opened 2018-08-05
add parameter language controlling tesseract model

bertsky:
allow passing tessdata model basename to parameter.json

### DIFF
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -13,6 +13,10 @@
           "type": "string",
           "enum": ["page", "region", "line", "word", "glyph"],
           "default": "line"
+        },
+        "language": {
+          "type": "string",
+          "description": "tessdata model to apply (an ISO 639-3 language specification or some other basename, e.g. deu-frak or Fraktur)"
         }
       }
     },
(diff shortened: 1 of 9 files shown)


ocrd_tesserocr/ocrd-tool.json
```diff
@@ -13,6 +13,10 @@
           "type": "string",
           "enum": ["page", "region", "line", "word", "glyph"],
           "default": "line"
+        },
+        "language": {
```
stweil: Perhaps using `model` instead of `language` would fit better. In the past, the model described a language, but recently "scripts" where added, so there exist now `script/Latin` or `script/Fraktur`, for example.

For OCR-D,  I expect that we will finally have several models (perhaps for different scripts or for the different centuries), but all will be for the same language.

stweil: We might also have some (rare) cases where more than one model should be passed.

bertsky:
looks like ci tests fail for several reasons older than this:
- OCR-D/core#154
- _test/test_recognize.py_ still uses older API for `Resolver` (vanished attribute `cache_enabled`)

stweil:
CC @noahmetzger – can you fix the ci tests?

kba:
IMHO it would be best to use URL of models, ideally in combination with checksums, to avoid any ambiguities (different versions of tesseract, best vs. fast models etc). What we're lacking at the moment is a ready-to-use model repository though.

Apart from that, I'd approve, I'll have a look at those unit tests later and feed back then.

PR #17, opened 2018-08-12
support more textequiv levels

bertsky:
This is an attempt to implement the other annotation levels. In my opinion, the behaviour for the different levels cannot be made completely analogous with Tesseract: simply pointing it to rectangles for words and glyphs (from an external layout segmentation) would produce results of far worse quality than always recognizing one complete line and allowing its own segmentation below it (accessible by iterators). In contrast, from the line level upwards we can reliably use its respective page segmentation mode (SINGLE_LINE / SINGLE_BLOCK / AUTO). 
Perhaps warnings and exceptions should be dealt with in a different, more systematic way though.

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -268,7 +268,7 @@ class TesserocrRecognize(Processor):
                             else:
                                 result_it.Next(RIL.WORD)
                 ID = concat_padded(self.output_file_grp, n)
-                self.add_output_file(
+                self.workspace.add_file(
                     ID=ID,
                     file_grp=self.output_file_grp,
                     basename=ID + '.xml',
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -113,7 +113,7 @@ class TesserocrRecognize(Processor):
                         tessapi.SetPageSegMode(PSM.SINGLE_BLOCK)
                         if region.get_TextEquiv():
                             log.warn("Region '%s' already contains text results", region.id)
-                        region.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text()))
+                        region.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text().strip()))
                         continue # next region (to avoid indentation below)
                     ## line, word, or glyph level:
                     textlines = region.get_TextLine()
@@ -132,7 +132,7 @@ class TesserocrRecognize(Processor):
                         for (word_no, conf) in enumerate(tessapi.AllWordConfidences()): # or tessapi.MeanTextConf()?
                             line_conf *= conf/100.0
                         # add line annotation unconditionally (i.e. even for word or glyph level):
-                        line.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text(), conf=line_conf))
+                        line.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text().strip(), conf=line_conf))
                         if maxlevel == 'line':
                             # maybe add TextEquiv alternatives via ChoiceIterator for TEXTLINE?
                             continue # next line (to avoid indentation below)
@@ -150,7 +150,7 @@ class TesserocrRecognize(Processor):
                                 if word.get_TextEquiv():
                                     log.warn("Word '%s' already contains text results", word.id)
                                 word_conf = tessapi.AllWordConfidences()[0]/100.0
-                                word.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text(), conf=word_conf))
+                                word.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text().strip(), conf=word_conf))
                                 if maxlevel == 'word':
                                     # maybe add TextEquiv alternatives via ChoiceIterator for WORD?
                                     continue # next word (to avoid indentation below)
@@ -168,7 +168,7 @@ class TesserocrRecognize(Processor):
                                         if glyph.get_TextEquiv():
                                             log.warn("Glyph '%s' already contains text results", glyph.id)
                                         glyph_conf = tessapi.AllWordConfidences()[0]/100.0
-                                        glyph.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text(), conf=glyph_conf))
+                                        glyph.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text().strip(), conf=glyph_conf))
                                         # maybe add TextEquiv alternatives via ChoiceIterator for SYMBOL?
                                     continue # next word (to avoid indentation below)
                                 ## internal glyph layout:
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -1,8 +1,9 @@
 from __future__ import absolute_import
 
-from tesserocr import RIL, PSM, PyTessBaseAPI, get_languages, iterate_level
-from ocrd.utils import getLogger, concat_padded, xywh_from_points, points_from_x0y0x1y1
-from ocrd.model.ocrd_page import from_file, to_xml, TextEquivType, CoordsType, GlyphType
+from tesserocr import RIL, PSM, PyTessBaseAPI, PyResultIterator, get_languages, iterate_level
+from ocrd.utils import getLogger, concat_padded, xywh_from_points, points_from_xywh, points_from_x0y0x1y1
+from ocrd.model.ocrd_page import from_file, to_xml, TextEquivType, CoordsType, GlyphType, WordType
+from ocrd.model.ocrd_page_generateds import TextStyleType
 from ocrd import Processor, MIMETYPE_PAGE
 from ocrd_tesserocr.config import TESSDATA_PREFIX, OCRD_TOOL
 
@@ -20,8 +21,6 @@ class TesserocrRecognize(Processor):
         Performs the (text) recognition.
         """
         print(self.parameter)
-        if self.parameter['textequiv_level'] not in ['line', 'glyph']:
-            raise Exception("currently only implemented at the line/glyph level")
         model = get_languages()[1][-1] # last installed model
         if 'model' in self.parameter:
             model = self.parameter['model']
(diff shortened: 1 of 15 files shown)


kba:
> simply pointing it to rectangles for words and glyphs (from an external layout segmentation) would produce results of far worse quality than always recognizing one complete line and allowing its own segmentation below it (accessible by iterators).

Allowing recognition on sub-line segmentation was to make the hierarchy of recognition levels consistent, not to be practical. The applications I had in mind were for comparing reconition results without differing (word) segmentation interfering.

I am not sure whether to stop processing if existing word level recognition could contradict line level textequiv. I see your point but maybe it would be good to empty the line before doing recognition, optional via parameter?

Anyway, this looks good to me. Maybe @noahmetzger can weigh in on this PR?

bertsky:
Oh, I see. Yes, comparing Word level results without segmentation interfering seems like a valid use case to me now.

If we want to allow this, the inner loops will have to be rewritten more radically, though. We have to set `PSM.SINGLE_WORD` rather than erase something. (I had this as an intermediate and decided against it.) Nevertheless, one should still add a warning saying Tesseract cannot perform optimally on that input. Want me to  modify the PR along those lines?

noahmetzger:
I am working on a different approach for the same problem at the moment. I add a feature  in the API of tesseract to get glyph confidences and alternative glyphs. Then the plan is to access this feature via python wrapper.

kba:
> Want me to modify the PR along those lines?

I am not sure, at moment I am not planning to use sub-line segmentation/recognition. I had hopen @noahmetzger might have an opinion :-) @wrznr Any thoughts?

> Then the plan is to access this feature via python wrapper.

I started extending tesserocr for the `GetGlyphConfidences` you did to core tesseract last week: https://github.com/kba/tesserocr/tree/glyph-confidences

> I am working on a different approach for the same problem at the moment.

Are you sure it's the same problem? This PR is about recognition below line level, not necessarily glyph confidences.

noahmetzger:
> Are you sure it's the same problem? This PR is about recognition below line level, not necessarily glyph confidences. 

Ah sorry, i thought this was the intent of this PR. Are there other use cases to go below line level?

> I am not sure, at moment I am not planning to use sub-line segmentation/recognition. I had hopen @noahmetzger might have an opinion :-) @wrznr Any thoughts?

hmmm no not really at the moment

> I started extending tesserocr for the GetGlyphConfidences you did to core tesseract last week: https://github.com/kba/tesserocr/tree/glyph-confidences

Ah great, thanks! Im currently working on options to get more accumulated results and only alternative glyphs per charakter, not per timestep.

bertsky:
Actually this PR is about all levels – besides the page level. We originally had line level, then we got glyph level. So now come region/block and word level. A use case for word level would be word-level postcorrection (CIS group). 

The specific question associated with this PR is whether it is acceptable to raise an exception when word or glyph level recognizer annotation is requested with word or glyph layout annotation already present. Obviously, recognition suffers when word/glyph rectangles are enforced externally, but this might be a good use case nevertheless, as @kba pointed out.

kba:
Bumping so we can move forward with merging this.

> The specific question associated with this PR is whether it is acceptable to raise an exception when word or glyph level recognizer annotation is requested with word or glyph layout annotation already present.

I think no errors should be raised but at the same time, it is much easier to just stop processing than to have a convention what to do in that case: emptying elements? Updating individual text nodes (how)).

So I'm for merging unless someone has a differing opinion.

bertsky:
I do, in fact. As mentioned earlier, I think emptying elements or other bailouts are unnecessary: We can use the same part of the API that is used above word level right now (`SetRectangle`) with `SetPageSegMode(PSM.SINGLE_WORD)` as well. This would simply add another `TextEquiv` there (perhaps with the same warning "already contains text results" as above). All that could still be conditioned on whether any `Word` already existed before, so we would get:
- Word (and Glyph) annotation based on `PSM.SINGLE_WORD` (or `PSM.SINGLE_CHAR`) in addition to existing annotation
- Word (and Glyph) annotation based on `GetIterator()` if empty before

bertsky:
I agree, the burden would shift from the producer (in dealing with exceptions) to the consumer (in interpreting `TextEquiv` sequences). But it need not merely be an ordering convention: we could use the `dataType` (and `dataTypeDetails`, or even `comments`) attribute for marking specific origins of textual annotation. (Unfortunately, the additional `UserDefined` introduced in 2018 does not apply here.)

Apart from that, I do not see any possible confusion or ambiguity. This refined proposal will never yield twice as many `Word` elements or `Glyph` elements, but really just as many `TextEquiv` elements at those levels as often _ocrd-tesserocr-recognize_ is processed. 

To recap: 
- Within each `TextLine`, if any `Word` existed before (from a previous layout annotation at the word level), we would (issue a warning and) re-use its coordinates via `SetRectangle` and enter `PSM.SINGLE_WORD`, instead of raising an exception. 
- And – completely parallel to that – within each `Word` thus entered, if any `Glyph` existed before (from a previous layout annotation at the glyph level), we would (issue a warning and) re-use its coordinates via `SetRectangle` and enter `PSM.SINGLE_CHAR`, instead of raising yet another exception.

So, if you still find this use case valid, and supporting it by shifting to a mere warning for the producer and additional text annotation for the consumer, just let me know and I will modify the PR accordingly.

kba:
OK, I get it, we're discussing only recognition, not ambiguous segmentations. Creating new child elements if none exist yet but otherwise augment existing elements.

> So, if you still find this use case valid, and supporting it by shifting to a mere warning for the producer and additional text annotation for the consumer, just let me know and I will modify the PR accordingly.

👍 Sounds like the best way forward. This behavior can still be adapted later via parameters but it is a reasonable default.

bertsky:
Done. Also fixes #18. I am not so sure about the new `MetadataItem` but it is better than nothing I think.

bertsky:
No idea what causes that segfault in Travis' Python 3.4 run of `make test`. I cannot reproduce (with 3.6). Can we control the level of verbosity in the log?

Note: the segfault with 3.4 already happened before I un-skipped and extended _test_recognize.py_ so the problem must be related to Tesseract layout analysis or even (though unlikely) _core_ stuff.

ocrd_tesserocr/recognize.py
```diff
@@ -19,62 +24,251 @@ def process(self):
...
+            # user_words_file
+            # user_patterns_file
             for (n, input_file) in enumerate(self.input_files):
                 log.info("INPUT FILE %i / %s", n, input_file)
                 pcgts = from_file(self.workspace.download_file(input_file))
                 # TODO use binarized / gray
                 pil_image = self.workspace.resolve_image_as_pil(pcgts.get_Page().imageFilename)
                 tessapi.SetImage(pil_image)
+                metadata = pcgts.get_Metadata() # ensured by from_file()
+                metadata.add_MetadataItem(
```
kba: @VolkerHartmann Here's an example how parameter values could be stored in the PAGE instead of METS or externally.

ocrd_tesserocr/recognize.py
```diff
@@ -19,62 +24,251 @@ def process(self):
...
                 # TODO slow
                 #  tessapi.SetPageSegMode(PSM.SINGLE_LINE)
-                log.info("page %s", pcgts)
-                for region in pcgts.get_Page().get_TextRegion():
+                log.info("Recognizing text in page '%s'", pcgts.get_pcGtsId())
+                page = pcgts.get_Page()
+                if maxlevel == 'page':
+                    # not sure what to do here: 
+                    # - We cannot simply do GetUTF8Text(), because there is no TextEquiv on the page level.
+                    # - We could GetComponentImages(RIL.BLOCK) and add a text region for each, then enter region level recognition below. But what if regions are already annotated? How to go about non-text blocks?
```
kba: If textequiv_level is meant as maxlevel (which is semantically cleaner), this should be an error for page as it is implemented.

ocrd_tesserocr/recognize.py
```diff
@@ -19,62 +24,251 @@ def process(self):
...
                         #  tessapi.G
-                        #  print(tessapi.AllWordConfidences())
-                        if self.parameter['textequiv_level'] == 'glyph':
-                            for word in line.get_Word():
-                                log.debug("Recognizing text in word '%s'", word.id)
-                                xywh = xywh_from_points(word.get_Coords().points)
-                                tessapi.SetRectangle(xywh['x'], xywh['y'], xywh['w'], xywh['h'])
+                        line_conf = 1.0
+                        for (word_no, conf) in enumerate(tessapi.AllWordConfidences()): # or tessapi.MeanTextConf()?
+                            line_conf *= conf/100.0
```
kba: Is that specified somewhere, that line confidence is the product of word confidences?

kba:
The CI failure is because of tesseract not a change in the PR (well you reenabled a failing test) or in core. It happened before. I will try to fix the Travis setup and/or use another CI service with a more recent OS version. Building tesseract for every run and every Python version is crazy anyway.

kba:
And I've tested it against 27, 34 and 36 and all is well.

ocrd_tesserocr/recognize.py
bertsky: True, although arguably "page" is invalid even if `textequiv_level` was not meant as maxlevel but "level above which to start recognition". (Maybe I am missing possible interpretations here.) Perhaps we should take "page" off the list in `ocrd_tool.json`?

kba:
Yes we should.

On August 29, 2018 9:55:37 PM GMT+02:00, Robert Sachunsky <notifications@github.com> wrote:
>bertsky commented on this pull request.
>
>
>
>>                  # TODO slow
>                 #  tessapi.SetPageSegMode(PSM.SINGLE_LINE)
>-                log.info("page %s", pcgts)
>-                for region in pcgts.get_Page().get_TextRegion():
>+                log.info("Recognizing text in page '%s'",
>pcgts.get_pcGtsId())
>+                page = pcgts.get_Page()
>+                if maxlevel == 'page':
>+                    # not sure what to do here: 
>+                    # - We cannot simply do GetUTF8Text(), because
>there is no TextEquiv on the page level.
>+                    # - We could GetComponentImages(RIL.BLOCK) and add
>a text region for each, then enter region level recognition below. But
>what if regions are already annotated? How to go about non-text blocks?
>
>True, although arguably "page" is invalid even if `textequiv_level` was
>not meant as maxlevel but "level above which to start recognition".
>(Maybe I am missing possible interpretations here.) Perhaps we should
>take "page" off the list in `ocrd_tool.json`?
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/OCR-D/ocrd_tesserocr/pull/17#discussion_r213813163

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ocrd_tesserocr/recognize.py
bertsky: No, not really. For true probabilities that would be correct I think. But apart from the percentage scale, these numbers need not sum up to 1 (or 100, respectively). This is what I see with Tesseract 3 models. For LSTM models it looks more like a probability, but that may change soon. So shall we prefer `MeanTextConf`?

bertsky:
Regarding the crazyness of building Tesseract in Travis: Can't we move towards the official ubuntu 18 repo versions now? These are named `4.00~git2288-10f499` (after [git rev 10f499](https://github.com/tesseract-ocr/tesseract/commit/10f4998aee3ccc68e9c4931ce744dd292ad6ff19) from Apr 2nd and tagged `4.0.0-beta.2`) and report "--version" as `4.0.0-beta.1` for tesseract and `1.75.3` for leptonica. 

The current data packages are named `4.00~git24-0e00fe6` (after [git rev 0e00fe6](https://github.com/tesseract-ocr/tessdata_fast/commit/0e00fe67ae71ead6155ea03ef13f1a9e77dd7093) from Feb 26th) are from *tessdata_fast* and probably not what we want.

kba:
Travis runs ubuntu 14.04 and I tried various solutions (alex' ppa, prebuilt debs, source) but something always broke, check out the build history. But ill investigate, i had better results on circleci with tesseract.

PR #24, opened 2018-09-26
override locale to POSIX before importing tesserocr

### LINKED ISSUE #23: new locale assertions in Tesseract are incompatible with Click
Ever since Tesseract 4 had to [introduce an assertion](https://github.com/tesseract-ocr/tesseract/issues/1670) that localization be plain POSIX (`C`) to ensure certain legacy assumptions in its code are always met, we have to override the current locale before initializing `tesserocr` API, too. This cannot be done by the user before calling any `ocrd_tesserocr` CLI, because we  depend on the Click library, which itself is [incompatible](http://click.palletsprojects.com/en/7.x/python3/#python-3-surrogate-handling) (in Python 3) with that locale (it requires at least `C.UTF-8`). So we have a deadlock.

We could perhaps reset the locale _after_ `click` and _before_ `tesserocr` though.

bertsky:
Fixes #23.

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -1,5 +1,8 @@
 from __future__ import absolute_import
 
+import locale
+locale.setlocale(locale.LC_ALL, 'C') # circumvent tesseract-ocr issue 1670 (which cannot be done on command line because Click requires an UTF-8 locale in Python 3)
+
 from tesserocr import RIL, PSM, PyTessBaseAPI, PyResultIterator, get_languages, iterate_level
 from ocrd.utils import getLogger, concat_padded, xywh_from_points, points_from_xywh, points_from_x0y0x1y1
 from ocrd.model.ocrd_page import from_file, to_xml, TextEquivType, CoordsType, GlyphType, WordType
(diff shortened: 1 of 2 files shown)


ocrd_tesserocr/recognize.py
```diff
@@ -33,6 +36,7 @@ def process(self):
                 raise Exception("configured model " + model + " is not installed")
         with PyTessBaseAPI(path=TESSDATA_PREFIX, lang=model) as tessapi:
             log.info("Using model '%s' in %s for recognition at the %s level", model, get_languages()[0], maxlevel)
+            tessapi.SetVariable("glyph_confidences", "2") # populate GetChoiceIterator() with LSTM models, too
```
kba: Interesting that tesseract doesn't validate these apparently because tests passed for older versions as well.

kba:
Thank you

ocrd_tesserocr/recognize.py
bertsky: Yes, I was going to ask Noah or Stefan about them. (As far as I understand, many of them are out-dated anyway. Since Tesseract needs to be downwards-compatible for applications, I guess the variables must be dragged along without further checking.)

PR #25, opened 2018-10-25
Refactor recognize

kba:
`process` has been too long and with too many nested loops and code duplication for the glyph-in-word. Not supposed to change any behavior.

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -103,176 +103,15 @@ class TesserocrRecognize(Processor):
                 log.info("Recognizing text in page '%s'", pcgts.get_pcGtsId())
                 page = pcgts.get_Page()
                 if maxlevel == 'page':
-                    # not sure what to do here: 
+                    # not sure what to do here:
                     # - We cannot simply do GetUTF8Text(), because there is no TextEquiv on the page level.
                     # - We could GetComponentImages(RIL.BLOCK) and add a text region for each, then enter region level recognition below. But what if regions are already annotated? How to go about non-text blocks?
                     raise Exception("currently only implemented below the page level")
                 ## region, line, word, or glyph level:
                 regions = page.get_TextRegion()
                 if not regions:
-                    log.warn("Page contains no text regions") 
-                for region in regions:
-                    log.debug("Recognizing text in region '%s'", region.id)
-                    if maxlevel == 'region':
-                        region_xywh = xywh_from_points(region.get_Coords().points)
-                        tessapi.SetRectangle(region_xywh['x'], region_xywh['y'], region_xywh['w'], region_xywh['h'])
-                        tessapi.SetPageSegMode(PSM.SINGLE_BLOCK)
-                        if region.get_TextEquiv():
-                            log.warn("Region '%s' already contains text results", region.id)
-                        region.add_TextEquiv(TextEquivType(Unicode=tessapi.GetUTF8Text().rstrip("\n\f")))
-                        continue # next region (to avoid indentation below)
-                    ## line, word, or glyph level:
-                    textlines = region.get_TextLine()
-                    if not textlines:
-                        log.warn("Region '%s' contains no text lines", region.id)
-                    for line in textlines:
-                        log.debug("Recognizing text in line '%s'", line.id)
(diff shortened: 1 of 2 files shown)


kba:
@bertsky Pleae have a look to make sure I didn't break any of it. Thanks

ocrd_tesserocr/recognize.py
```diff
@@ -1,15 +1,16 @@
...
-from tesserocr import RIL, PSM, PyTessBaseAPI, PyResultIterator, get_languages, iterate_level
-from ocrd.utils import getLogger, concat_padded, xywh_from_points, points_from_xywh, points_from_x0y0x1y1
+from tesserocr import RIL, PSM, PyTessBaseAPI, get_languages
+from ocrd.utils import getLogger, concat_padded, xywh_from_points, points_from_x0y0x1y1
 from ocrd.model.ocrd_page import from_file, to_xml, TextEquivType, CoordsType, GlyphType, WordType
 from ocrd.model.ocrd_page_generateds import TextStyleType, MetadataItemType, LabelsType, LabelType
 from ocrd import Processor, MIMETYPE_PAGE
 from ocrd_tesserocr.config import TESSDATA_PREFIX, OCRD_TOOL
 
+locale.setlocale(locale.LC_ALL, 'C') # circumvent tesseract-ocr issue 1670 (which cannot be done on command line because Click requires an UTF-8 locale in Python 3)
```
bertsky: I am afraid this will break #23 again. The locale needs to be reset **before** importing from tesserocr/Tesseract the first time.

ocrd_tesserocr/recognize.py
```diff
@@ -281,3 +120,157 @@ def process(self):
...
+            if maxlevel == 'word':
+                # maybe add TextEquiv alternatives via ChoiceIterator for WORD?
+                continue # next word (to avoid indentation below)
+            ## glyph level:
+            glyphs = word.get_Glyph()
+            if glyphs:
+                ## external glyph layout:
+                # raise Exception("existing annotation for Glyph level would clash with OCR results for word '%s'", word.id) # forcing external layout annotation for gylphs is worse with Tesseract
+                log.warning("Word '%s' contains glyphs already, recognition might be suboptimal", word.id)
+                for glyph in glyphs:
```
bertsky: I really like the idea of how you split up `self.process` into levels and "existing" (external) vs non-existing (internal) structure. Maybe for consistency and clarity, at this point one could likewise separate #L233 through #L248 into a `self._process_existing_glyphs(glyphs, tessapi)`?

ocrd_tesserocr/recognize.py
kba: Oh right. Should test CLI call as well.

kba: Added a simple test, running the CLI and [it passed](https://travis-ci.org/OCR-D/ocrd_tesserocr/jobs/446218522). Won't the locale problem only arise once PyTessAPI is initialized?

kba: Nope, that was too quick, fails indeed for Python 3.

ocrd_tesserocr/recognize.py
kba: Good idea, done.

ocrd_tesserocr/recognize.py
kba: OK, rearranged it and tests pass for py3+tess4.

bertsky: I cannot see that fix yet. Can you please push?

kba: https://github.com/OCR-D/ocrd_tesserocr/commit/cc6b56358f25b79443a009cf87aa62ad16010a8b

bertsky: > cc6b563

That's strange. I can see it here on the web server, but still nothing is available via `git fetch upstream pull/25/head` (where "upstream" is `ssh://git@github.com/OCR-D/ocrd_tesserocr`). Is that normal for GH?

bertsky: sorry, I just forgot to merge FETCH_HEAD (or use pull) for the follow-up changes.

Issue #26, opened 2018-12-06
File names


finkf:
There are three files `{0006,0007,0008}.xml` that all belong to the same filegroup `gt`. If I run `ocrd-tesserocr-recognize` on the filegroup `gt`, with output filegroup `tess` recognize searches for the files of the filegroup in the `mets.xml` file. If for some reason (files where not added to the workspace in nummerical order?) the files are not returned in numerical order - for example 0007, 0008, 0006 - recognize generates the files `tess-0001.xml` (0007.xml), `tess-0002.xml` (0008.xml) and `tess-0003.xml` (0006.xml). 

This destroys the mapping between gt and ocr pages. 

A simple solution would be to use:
```python
self.workspace.add_file(
  ID=ID,
  file_grp=self.output_file_grp,
  basename=self.output_file_grp + '-' + os.path.basename(input_file.url),
  mimetype=MIMETYPE_PAGE,
  content=to_xml(pcgts),
)
```
to create the new files to the workspace.

kba:
Sounds reasonable. I would change the `ID` though, so the mapping between filename and ID is retained.

finkf:
That's OK as well.

PR #27, opened 2018-12-06
more intuitive ID for output file, #26

kba:
Generate the output ID and filename from the input file ID reduced to its numbers.

@finkf


finkf:
or maybe even better:

```python
ID = concat_padded(self.output_file_grp, os.path.basename(input_file.url)[:-4])
```
why generate ids if output_file_grp + basename of file without extension is unique?

bertsky:
I am very much in favour of the solution by @finkf, but I would also like to keep the `.xml` extension in the old version (because most PAGE viewers rely on it). The patch does not apply anymore, so should I make a new PR?

bertsky:
Closing as this is superseded (and hopefully resolved to satisfaction) by #48.

PR #29, opened 2019-02-14
repair automated build and tests:

bertsky:
- fix makefile and requirements:
  - remove patch-header (not applicable anymore)
  - reinstate ubuntu deps, also for tesseract itself
  - avoid tesserocr 2.4.0 because tesseract 4.0.0
    is not available via apt yet

- fix dockerfile:
  - remove patch-header (not applicable anymore)
  - do not install core (already in base container)
  - install tesserocr via pip instead of repo
  - make sure makefile recipes are run with python3
  - avoid copying cached/repo files
  - also install test environment

- fix tests:
  - update test-cli to new data/ convention
  - update test to renamed paths in assets repo

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,7 +2,12 @@
 !Makefile
 !setup.py
 !requirements.txt
+!requirements_test.txt
 !LICENSE
 !README.rst
 
-!ocrd_tesserocr
+# avoid .git and __pycache__ etc:
+!ocrd_tesserocr/**/*.py
+!ocrd_tesserocr/**/*.json
+!test/**/*.py
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ ENV LANG C.UTF-8
 WORKDIR /build-ocrd
 COPY setup.py .
 COPY requirements.txt .
+COPY requirements_test.txt .
 COPY README.rst .
 COPY LICENSE .
 RUN apt-get update && \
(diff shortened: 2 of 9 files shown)


bertsky:
note: the Dockerfile did not work at all before, so apt install is better than nothing. Travis merely failed because of changes in assets. 

We should be able to keep our Travis running until we find a way to base it on 4.0.0 as well. (Can we migrate from Ubuntu 14 to 18?) The sooner the better – we will need the 4.0.0 features soon.

bertsky:
argh, now the workspace download fails for the recognizer test in Travis – this is new. Maybe related to OCR-D/core#96?

bertsky:
Just fixed the `dst_dir` in 2 of the 4 tests, but the above mentioned resolver/workspace problem remains.

bertsky:
So 0de1d85 changes behaviour significantly – please review! (I could place it in a separate PR if required.)

kba:
> We should be able to keep our Travis running until we find a way to base it on 4.0.0 as well. (Can we migrate from Ubuntu 14 to 18?) The sooner the better – we will need the 4.0.0 features soon.

No, travis won't. But we can use CircleCI instead which uses Docker and gives greater freedom for environment.

> argh, now the workspace download fails for the recognizer test in Travis – this is new. Maybe related to [OCR-D/core#96](https://github.com/OCR-D/core/issues/96)?

This needs fixing and fix should (hopefully) be in before 2019-02-27.

I'll merge this now, so I can test interop with upcoming 1.0 release. Thanks for the effort.

PR #34, opened 2019-04-17
[WIP] Add deskewing per tesseract

wrznr:
With the help of `OSD`, tesseract determines the skew angle for
images. The wrapper applies this to pages and regions. It is
not clear yet how to save the estimated skew angle in PAGE
XML. Cf. https://github.com/PRImA-Research-Lab/PAGE-XML/issues/9

### DIFF
--- a/ocrd_tesserocr/deskew.py
+++ b/ocrd_tesserocr/deskew.py
@@ -1,0 +1,91 @@
+from __future__ import absolute_import
+
+import locale
+
+# pylint: disable=wrong-import-position
+locale.setlocale(locale.LC_ALL, 'C') # circumvent tesseract-ocr issue 1670 (which cannot be done on command line because Click requires an UTF-8 locale in Python 3)
+
+from tesserocr import RIL, PSM, PyTessBaseAPI
+
+from ocrd_utils import getLogger, concat_padded, xywh_from_points, points_from_x0y0x1y1, MIMETYPE_PAGE
+from ocrd_modelfactory import page_from_file
+from ocrd_models.ocrd_page import (
+    CoordsType,
+    GlyphType,
+    LabelType,
+    LabelsType,
+    MetadataItemType,
+    TextEquivType,
+    TextStyleType,
+
+    to_xml
+)
+from ocrd import Processor
+from .config import TESSDATA_PREFIX, OCRD_TOOL
+
+log = getLogger('processor.TesserocrDeskew')
+
+class TesserocrDeskew(Processor):
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


ocrd_tesserocr/deskew.py
```diff
@@ -0,0 +1,91 @@
...
+        # print(self.parameter)
+        oplevel = self.parameter['operation_level']
+        with PyTessBaseAPI(path=TESSDATA_PREFIX, psm=PSM.AUTO_OSD) as tessapi:
+            for (n, input_file) in enumerate(self.input_files):
+                log.info("INPUT FILE %i / %s", n, input_file)
+                pcgts = page_from_file(self.workspace.download_file(input_file))
+                pil_image = self.workspace.resolve_image_as_pil(pcgts.get_Page().imageFilename)
+
+                metadata = pcgts.get_Metadata() # ensured by from_file()
+                metadata.add_MetadataItem(
```
wrznr: I copied the `metadata` code from https://github.com/OCR-D/ocrd_tesserocr/blob/d968abcd62dab1f862faa7df4e1676d70322402f/ocrd_tesserocr/recognize.py#L106 ff. No idea what it actually does or if it is necessary. In addition, `segment_region` does not have it.

wrznr:
@bertsky This means: If I iterate over the blocks and determine their skew, the first block will yield an identical value? I will check this. If it tuns out to be the case, I will restrict the operation on blocks, okay?

bertsky:
@wrznr Yes exactly, that what I would expect. Yes, `RIL.BLOCK` should be good (via simple for loop on `iterate_level` or using `Next` and `IsAtFinalElement`, as in recognize.py).

wrznr:
`_process_regions` is already there!

bertsky:
> `_process_regions` is already there!

Right, this may work, too. (I still had on my mind what the API suggests: using the blocks from its PageIterator, not from an external layout analysis. We had a similar situation with text recognition BTW: results may be worse if the engine does not get to see the kinds of layout it was trained on. Maybe one should try to compare both modes empirically, just to be sure how robust this is?)

wrznr:
@bertsky Is it necessary to update this PR or do your forthcoming additions to ocrd_tesserocr contain this proposed functionalities?

bertsky:
@wrznr I started out independently, but I could of course rebase on this PR and either add or squash and force-push. It keeps making more sense by the minute :-)

bertsky:
Sorry, I rebased this on the current master, but was not subsequently allowed to force-push here. So I started a new PR instead (#48), which supersedes this one.

PR #35, opened 2019-05-04
improve consistency and quality of results

bertsky:


ocrd_tesserocr/ocrd-tool.json
```diff
@@ -24,6 +24,11 @@
           "default": "line",
           "description": "PAGE XML hierarchy level to add the TextEquiv results to (requires existing layout annotation up to one level above that)"
         },
+        "overwrite_lines": {
+          "type": "boolean",
+          "default": false,
+          "description": "remove existing layout and text annotation below the TextLine level"
```
wrznr: If it is "below" the text line level, pls. choose a different name, `overwrite_lines` suggests a new segmentation of regions into lines.

bertsky: Oh right, that did not occur to me. (I was more concerned about confusion with `textequiv_level`. This parameter is tied to the line level because the OCR itself is internally.) How about `overwrite_words`, despite the fact that glyphs will be removed as well? Or would `overwrite_line` be unambiguous enough?

wrznr: Both okay for me with a slight preference for `overwrite_words`.

bertsky:
> Looks good to me. Have you filed an issue concerning the PAGE XML inconsistency?

No, since this started out as a "feature" (allowing different processing levels in the same document). Should I open one and refer here, so this can be found more easily?

bertsky:
Also, I should probably define a new version and append the changelog.

bertsky:
Sorry about all the force-pushing noise! I have no other way to test the Travis build, the Docker image deviates significantly.

bertsky:
Argh, I just don't get why in Travis the locale assertion keeps firing. Is there any way to get into that container interactively for debugging?

wrznr:
I do not think so. @kba?

bertsky:
I will try [this answer on SO](https://stackoverflow.com/a/38804734) ...

wrznr:
I do not see an assert in https://github.com/tesseract-ocr/tesseract/blob/af3dd1af066e6ed37ba6a4b437c5565a4eceaf95/src/api/baseapi.cpp#L209

bertsky:
You are right, it's gone! So @stweil has already succeeded. But for ocrd_tesserocr, we rely on prebuilt releases, and [here](https://github.com/nijel/tesseract-ocr-build/releases) the newest is 4.0.0...

stweil:
The assertions were removed in Git master and also in the Tesseract 4.1 branch. Hopefully there will be prebuilt 4.1 releases soon.

bertsky:
> I will try [this answer on SO](https://stackoverflow.com/a/38804734) ...

This does not work. (For one, the Ruby installation needs TLS 1.2, which requires a newer openssl / libssl-dev. To get it updated, the apt config must be repaired, because lots of the Ubuntu 14 repos are unavailable today. Next, I still had to work around [this issue](https://github.com/travis-ci/travis-ci/issues/8098). Now it will generate a ci.sh, but that does not run. I gave up after fixing [this issue](https://github.com/travis-ci/travis-ci/issues/3350) and running into many more.)

Does anyone have an idea how to proceed?

bertsky:
(There are many other changes waiting in my queue...)

bertsky:
@kba Can we move to a Docker-based CI now? This really keeps slowing down development too much. You mentioned CircleCI as a candidate earlier...

bertsky:
Anyway, switching to a different CI would be a separate PR. 

> Does anyone have an idea how to proceed?

Since those failing tests are a false positive, should I merge regardless?

kba:
Thanks for your work, looks great.

> Since those failing tests are a false positive, should I merge regardless?

Fine by me and tests pass locally as well. Agreed, Travis is based on tech so outdated now that it's getting hard to work with. I will investigate CircleCI.

bertsky:
Thanks!

bertsky:
Oh, forgot to mention: fixes #38, fixes #39.

PR #41, opened 2019-06-17
CircleCI

kba:
Replace Travis CI with Circle CI based on Ubuntu 18.04 docker container.


bertsky:
Thanks @kba!

Unfortunately, I cannot get `circleci-cli local execute --job build-python36` running. It complains there is no `git` installed – right in the `checkout` step. I tried adding a command step for `apt-get install -y git` before that, but then it cannot locate any packages – and checkout must be first according to the docs. I have also tried adding a second container image `circleci/python:3.6`, but that does not change a thing.

The problem seems to be that the vanilla `ubuntu` docker image does not have git installed, only the `circleci/*` images do. In the server/non-local setup, the log clearly states:
> Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.

So, I guess circleci-cli is not an option for us, but we can now trust that `docker run ocrd/tesserocr "make PYTHON=python3 PIP=pip3 test"` will yield the same result like the CI test from now on, right?

kba:
I cannot run circleci here because of firewall woes (apt-get connections time out)

> I tried adding a command step for apt-get install -y git before that, but then it cannot locate any packages

Did you run `apt-get update` before `apt-get install -y git`?

> `docker run ocrd/tesserocr "make PYTHON=python3 PIP=pip3 test"`

Again cannot test it here for the same reason (docker network not bridged through or something). But since ocrd/tesserocr is based on ocrd/core, which in turn is based on ubuntu:18.04 **yes** this should have the same behavior.

I have to revisit the docker setup from another location soon. Would probably make sense to base even the CI docker images on ocrd/core to make the process even more consistent.

bertsky:
> Did you run `apt-get update` before `apt-get install -y git`?

No. But if I do, then it **does** work with circleci-cli!

> Would probably make sense to base even the CI docker images on ocrd/core to make the process even more consistent.

Why not just the `ocrd/tesserocr` docker image? This should be way faster, and even more consistent, no?

(Strangely, if I do this, replacing ubuntu with the ocrd/tesserocr image, I end up with all tests failing because Python cannot import `ocrd.resolver` – although ocrd==1.0.0b8 is installed, and despite getting all tests succeeding via docker interactively.)

kba:
> Why not just the ocrd/tesserocr docker image? This should be way faster, and even more consistent, no?

Would be possible but I find that confusing since for testing I don't want any older duplicates of the same code in the container.

Why it actually fails I cannot say. Automatic building of docker containers on docker.hub is patchy apparently, ocrd-tesserocr was last built 2 months ago, there are no autobuilds for ocrd/core at all apparently. Need to investigate this further from a less restricted network, c.f. https://github.com/OCR-D/ocrd_tesserocr/issues/44

bertsky:
I see. Thanks for explaining!

BTW, I installed circleci-cli via [GH releases](https://github.com/CircleCI-Public/circleci-cli/releases) (just download+unpack) – I don't think there is an apt pkg for it anywhere.

kba:
The apt-get errors are from docker, installing circleci-cli works fine.

bertsky:
> The apt-get errors are from docker, installing circleci-cli works fine.

Oh, now I get it.

PR #54, opened 2019-07-09
Set DPI

bertsky:


bertsky:
I added the fix in `crop` which prevents negative coordinates when `AlternativeImage` was already present. Now also fixes #55.

bertsky:
Oh, @kba, am I supposed to use `OcrdExif` instead of just `PIL.Image.info`?

wrznr:
@bertsky We should choose a definite option for accessing image information in the long run. This could be either `exif` or `PIL`.

wrznr:
After a little research, I think that the image resolution is stored in JFIF meta data section which both `exiftool` and `PIL.Image.info` read:
```shell
$ exiftool IMG/FILE_0033_ORIGINAL.jpg 
ExifTool Version Number         : 10.80
File Name                       : FILE_0033_ORIGINAL.jpg
Directory                       : IMG
File Size                       : 329 kB
File Modification Date/Time     : 2019:07:04 17:07:41+02:00
File Access Date/Time           : 2019:07:09 14:40:50+02:00
File Inode Change Date/Time     : 2019:07:04 17:07:46+02:00
File Permissions                : rw-r--r--
File Type                       : JPEG
File Type Extension             : jpg
MIME Type                       : image/jpeg
JFIF Version                    : 1.01
Resolution Unit                 : inches
X Resolution                    : 300
Y Resolution                    : 300
Image Width                     : 1187
Image Height                    : 1687
Encoding Process                : Baseline DCT, Huffman coding
Bits Per Sample                 : 8
Color Components                : 3
Y Cb Cr Sub Sampling            : YCbCr4:2:0 (2 2)
Image Size                      : 1187x1687
Megapixels                      : 2.0
```
vs.
```python
>>> from PIL import Image
>>> Image.open("IMG/FILE_0033_ORIGINAL.jpg")
<PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=1187x1687 at 0x7F7958245C18>
>>> img = Image.open("IMG/FILE_0033_ORIGINAL.jpg")
>>> img.info
{'jfif': 257, 'jfif_version': (1, 1), 'dpi': (300, 300), 'jfif_unit': 1, 'jfif_density': (300, 300)}
```

bertsky:
That's right, but there are more options than JPG here. And it can always be empty. And core's `OcrdExif` _looks_ like a good abstraction (which does use PIL already, btw). That's why I was asking whether this is recommended and stable (especially the fallback defaults `xResolution=1` etc). In that case it should probably be in the spec as well (with a hint that `ocrd.OcrdExif` already implements it).

kba:
>  That's why I was asking whether this is recommended and stable (especially the fallback defaults xResolution=1 etc)

This is the recommended way, though not as stable as I'd wish. The variability of image metadata is immense and `PIL.Image.Info` exposes the metadata on a relatively low level, so IMHO an abstraction is necessary to handle all the formats, units and bogus metadata.

The `xResolution = 1` hacks are in place not too cause division-by-zero errors and be so obviously low to prevent processors from assuming it to be true.

> In that case it should probably be in the spec as well (with a hint that ocrd.OcrdExif already implements it).

True, that should be made explicit. We have some information at https://ocr-d.github.io/gt//trans_documentation/exif_pagexml_konkordanz.html from early 2018 when we considered encoding image metadata with [MIX](http://www.loc.gov/standards/mix/).

ocrd_tesserocr/crop.py
```diff
@@ -146,7 +151,19 @@ def process(self):
                     LOG.debug("Padded page border: %i:%i,%i:%i", min_x, max_x, min_y, max_y)
                     border = BorderType(Coords=CoordsType(
                         points_from_bbox(min_x, min_y, max_x, max_y)))
+                    # update PAGE (annotate border):
                     page.set_Border(border)
+                    # update METS (add the image file):
+                    page_image = page_image.crop(
+                        box=(min_x, min_y, max_x, max_y))
+                    file_id = input_file.ID.replace(self.input_file_grp, FILEGRP_IMG)
```
kba: Could this cause duplicate ID for METS not following our conventions for identifiers?

ocrd_tesserocr/deskew.py
```diff
@@ -85,6 +85,9 @@ def process(self):
                                                                for name in self.parameter.keys()])]))
                 page = pcgts.get_Page()
                 page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
+                dpi = page_image.info.get('dpi', (0,0))[0]
```
kba: as above

ocrd_tesserocr/recognize.py
```diff
@@ -133,6 +133,9 @@ def process(self):
                                                                for name in self.parameter.keys()])]))
                 page = pcgts.get_Page()
                 page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
+                dpi = page_image.info.get('dpi', (0,0))[0]
```
kba: see above

ocrd_tesserocr/segment_line.py
```diff
@@ -71,6 +71,9 @@ def process(self):
                                                                for name in self.parameter.keys()])]))
                 page = pcgts.get_Page()
                 page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
+                dpi = page_image.info.get('dpi', (0,0))[0]
```
kba: see above

ocrd_tesserocr/segment_word.py
```diff
@@ -71,6 +71,9 @@ def process(self):
                                                                for name in self.parameter.keys()])]))
                 page = pcgts.get_Page()
                 page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
+                dpi = page_image.info.get('dpi', (0,0))[0]
```
kba: see above

ocrd_tesserocr/segment_region.py
```diff
@@ -126,6 +126,9 @@ def process(self):
                     else:
                         LOG.warning('keeping existing ReadingOrder')
                 page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
+                dpi = page_image.info.get('dpi', (0,0))[0]
```
kba: see above

ocrd_tesserocr/crop.py
```diff
@@ -87,8 +88,12 @@ def process(self):
                                 min_x, max_x, min_y, max_y)
                 
                 page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
-                dpi = page_image.info.get('dpi', (300,300))[0]
-                zoom = 300 / dpi
+                dpi = page_image.info.get('dpi', (0,0))[0]
```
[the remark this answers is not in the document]
bertsky: Thanks! I will add that to the PR as soon as my HDD is operable again.

> BTW: Should we offer some unified resolution derived from x- and y-Resolution?

With pleasure!

And we need to make the fallback (be it `1` or another) be decisive and stable, so we have a means to known when _not_ to override Tesseract's own DPI guessing.

ocrd_tesserocr/crop.py
bertsky: Absolutely true! I forgot to capture this here (as I did with the PAGE result `file_id` at the end of the processor).

bertsky:
> > In that case it should probably be in the spec as well (with a hint that ocrd.OcrdExif already implements it).
> 
> True, that should be made explicit.

But on the other hand, how should this possibly be related to the spec? We cannot teach the implementors how to talk to their underlying processors (in how they should determine and pass down DPI values). We can only make the _annotation_ in PAGE mandatory (i.e. `imageXResolution` / `imageYResolution` / `imageResolutionUnit`). 

Then we would of course generously provide a specialized processor that just annotates that (using `OcrdExif`). And we would of course build it into the basic `OcrdPage` functions as well, to add those attributes if they are not already there (for everyone who uses core to write their processors, so they do not need the special processor). Thus, any careful and wise processor author can find joy in just relying on the attributes of `PageType` for her DPI adjustments. She will also have to deal with the question of fallback/default values now: if nothing was annotated, then there is nothing to know, so she decides what to do about it herself.

ocrd_tesserocr/crop.py
bertsky: done

ocrd_tesserocr/crop.py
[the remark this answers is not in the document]
bertsky: So @kba I guess I will wait for you to release a new version of `OcrdExif`, or maybe even to ensure annotation in `Page` attributes, as outlined below? Until then this can stay the way it is, right?

kba:
> But on the other hand, how should this possibly be related to the spec? We cannot teach the implementors how to talk to their underlying processors (in how they should determine and pass down DPI values). We can only make the annotation in PAGE mandatory (i.e. imageXResolution / imageYResolution / imageResolutionUnit).

We had long discussions on this last year and decided to use the image itself as the single source of truth for all technical image metadata. Duplicating the information from the image in PAGE could lead to metadata drift (e.g. the image changes but the PAGE doesn't reflect that or the annotations are missing in PAGE) whereas the image is always available. In addition, the 2017 PAGE version didn't have the attributes yet.

How this is related to the spec: Different image file formats support different metadata schemes, like EXIF, JFIF or PNG tags. While an abstraction like `PIL.Image.Info`, it leaks pretty quickly. Implementing this to support the reality of images used in libraries, archives etc. is not only tedious for extracting metadata like dimension or pixel density but also for colorspace conversions or image manipulation.

Originally, the plan was to have an image characterization step that does all the metadata extraction, like you propose, and refuse processing bad images if pixel density too low etc. In reality, it is next to impossible to guess pixel density and the image metadata often cannot be trusted.

That being said, I'm open to encoding this in PAGE instead if that makes life easier for developers and users. This would require a dedicated processor as the first step in any pipeline though, since the input file group will be images.

@cneud @tboenig Any thoughts on this?

ocrd_tesserocr/crop.py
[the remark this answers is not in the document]
kba: > release a new version of OcrdExif

You mean with the unified resolution?

>  Until then this can stay the way it is, right?

I would really use OcrdExif even now because your code won't reliably handle JPEG or PNG resolution. But it's not essential for the PR here, can be changed later if it's not flexible enough.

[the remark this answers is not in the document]
bertsky: > You mean with the unified resolution?

Yes. And especially the stable (guaranteed) default.

> I would really use OcrdExif even now because your code won't reliably handle JPEG or PNG resolution. But it's not essential for the PR here, can be changed later if it's not flexible enough.

Okay, than I will change it now and we will already be able to gather experience with `OcrdExif`. I will only use `xResolution` and rely on `1` as being unset for now.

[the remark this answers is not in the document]
bertsky: Oh, and one more thing, @kba: It looks to me like the code which determines `resolutionUnit` for JPEG does not convert the internal JFIF integers to strings like for the other formats. See [here](https://github.com/python-pillow/Pillow/blob/1c57a41ececc4529520a423b9bec13c78927cb25/src/PIL/JpegImagePlugin.py#L86) for how Pillow creates `jfif_unit`.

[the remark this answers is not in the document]
bertsky: Confirmed! I will open an issue about it.

bertsky:
Related: OCR-D/core#37 (re-open there?)

ocrd_tesserocr/crop.py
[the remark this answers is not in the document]
bertsky: done: OCR-D/core#256

PR #56, opened 2019-07-14
make common functions process polygons instead of bboxes

bertsky:
...and improve readability + encapsulation

### DIFF
--- a/ocrd_tesserocr/common.py
+++ b/ocrd_tesserocr/common.py
@@ -5,7 +5,7 @@ import sys
 import io
 
 import numpy as np
-from PIL import Image, ImageDraw
+from PIL import Image, ImageDraw, ImageStat
 
 from ocrd_utils import getLogger, xywh_from_points, polygon_from_points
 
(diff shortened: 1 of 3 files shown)


bertsky:
This is pretty much the final version that could enter core. 

It just crossed my mind that now `image_from_line` is exactly the same as `image_from_word` and `image_from_glyph`, so they could be unified and renamed into some generalized `image_from_segment`, or exposed under distinct names but delegated to the unified implementation. In contrast, `image_from_region` still differs because regions can have `@orientation` themselves and their parents can (soon), too.

bertsky:
Wait, it gets better: We could even unify `image_from_region` if we look in the segment instance's attributes whether it has `get_orientation` and look into the `xywh` dict whether it has a key `angle`. That would even allow removing the FIXMEs waiting for adoption of [page-level @orientation](https://github.com/PRImA-Research-Lab/PAGE-XML/commit/2a1e5c706b7aa5cbd105cca124e19d740f64da08) in our PAGE library. 

And once we have that, I don't see any more reasons to not incorporate `resolve_image_as_pil(page.imageFilename)` into `image_from_page`, too.

That would leave processor designers with just 2 functions to remember, `image_from_page` and `image_from_segment` (instead of `resolve_image_as_pil`, `image_from_page`, `image_from_region`, `image_from_line`, `image_from_word`, `image_from_glyph`). Opinions?

ocrd_tesserocr/common.py
```diff
@@ -5,43 +5,145 @@
...
-    # point := (point - region_center) * region_rotation + region_center
-    # moreover, since rotation has reshaped/expanded the image,
-    # the line coordinates must be offset by those additional pixels:
-    # point := point + 0.5 * (new_region_size - old_region_size)
+def image_from_polygon(image, polygon):
+    """"Mask an image with a polygon.
+    
+    Given a PIL.Image `image` and a numpy array `polygon`
+    of relative coordinates into the image, fill everything
+    outside the polygon hull to the background. Since `image`
```
wrznr: *fill everything ... to the background*? Maybe *put* or *move* would be more appropriate?

ocrd_tesserocr/common.py
```diff
@@ -5,43 +5,145 @@
...
+    
+    Return a new PIL.Image.
+    """
+    mask = polygon_mask(image, polygon)
+    # create a background image from its median color
+    # (in case it has not been binarized yet):
+    # array = np.asarray(image)
+    # background = np.median(array, axis=[0, 1], keepdims=True)
+    # array = np.broadcast_to(background.astype(np.uint8), array.shape)
+    background = ImageStat.Stat(image).median[0]
```
wrznr: This could turn nasty in edge cases, right? Do we have use cases where non-binarized images are masked?

ocrd_tesserocr/common.py
```diff
@@ -5,43 +5,145 @@
...
+    white).
+    
+    Return a new PIL.Image.
+    """
+    # todo: perhaps we should issue a warning if we encounter this
+    # (It should be invalid in PAGE-XML to extend beyond parents.)
+    if not box:
+        box = (0, 0, image.width, image.height)
+    xywh = xywh_from_bbox(*box)
+    background = ImageStat.Stat(image).median[0]
```
wrznr: See above.

ocrd_tesserocr/common.py
```diff
@@ -5,43 +5,145 @@
...
+    new_image = Image.new(image.mode, (xywh['w'], xywh['h']),
```
wrznr: Is the median **defined** as white for binarized images?

ocrd_tesserocr/common.py
```diff
@@ -5,43 +5,145 @@
...
     # active rotation:  [[cos, -sin], [sin, cos]]
     # passive rotation: [[cos, sin], [-sin, cos]] (inverse)
-    return [(orig['x']
-             + (x - orig['x'])*np.cos(angle)
-             + (y - orig['y'])*np.sin(angle),
-             orig['y']
-             - (x - orig['x'])*np.sin(angle)
-             + (y - orig['y'])*np.cos(angle))
-            for x, y in coordinates]
+    return orig + np.dot(polygon - orig, np.array([[cos, sin], [-sin, cos]]).transpose())
```
wrznr: Real math! :+1:

ocrd_tesserocr/common.py
```diff
@@ -5,43 +5,145 @@
...
+    """
+    # get polygon:
+    polygon = np.array(polygon_from_points(segment.get_Coords().points))
+    # offset correction (shift coordinates to base of segment):
+    polygon -= np.array([parent_xywh['x'], parent_xywh['y']])
+    # angle correction (rotate coordinates if image has been rotated):
+    if 'angle' in parent_xywh:
+        polygon = rotate_coordinates(
+            polygon, parent_xywh['angle'],
+            orig=np.array([0.5 * parent_image.width,
```
wrznr: Pls. apologize: I do not understand the constant here...

ocrd_tesserocr/common.py
```diff
@@ -66,12 +172,18 @@ def image_from_page(workspace, page,
                  'y': 0,
                  'w': page_image.width,
                  'h': page_image.height}
+    # FIXME: uncomment as soon as we get @orientation in PageType:
```
wrznr: We should have this by now.

wrznr: @kba What is the status of the PAGE 2019 issue?

ocrd_tesserocr/common.py
```diff
@@ -66,12 +172,18 @@ def image_from_page(workspace, page,
...
+    # # region angle: PAGE orientation is defined clockwise,
+    # # whereas PIL/ndimage rotation is in mathematical direction:
+    # page_xywh['angle'] = -(page.get_orientation() or 0)
     # FIXME: remove PrintSpace here as soon as GT abides by the PAGE standard:
```
wrznr: @tboenig What is the status of the `Border` vs. `PrintSpace` issue?

ocrd_tesserocr/common.py
```diff
@@ -83,16 +195,25 @@ def image_from_page(workspace, page,
         page_image = workspace.resolve_image_as_pil(
             alternative_image[-1].get_filename())
     elif border:
```
wrznr: I do not fully understand the logistics here: What if we have `border` without `Coords`? (See `if` condition above).

ocrd_tesserocr/common.py
```diff
@@ -83,16 +195,25 @@ def image_from_page(workspace, page,
...
+        page_image = image_from_polygon(page_image, page_polygon)
+        # recrop into page rectangle:
+        page_image = crop_image(page_image,
             box=(page_xywh['x'],
                  page_xywh['y'],
                  page_xywh['x'] + page_xywh['w'],
                  page_xywh['y'] + page_xywh['h']))
-        # FIXME: mask away all GraphicRegion, SeparatorRegion etc which
-        # could overlay any text regions
+        # FIXME: uncomment as soon as we get @orientation in PageType:
```
wrznr: See above.

ocrd_tesserocr/common.py
```diff
@@ -83,16 +195,25 @@ def image_from_page(workspace, page,
...
+        #     LOG.info("About to rotate page '%s' by %.2f°",
+        #               page_id, page_xywh['angle'])
+        #     page_image = page_image.rotate(page_xywh['angle'],
+        #                                        expand=True,
+        #                                        #resample=Image.BILINEAR,
+        #                                        fillcolor='white')
     # subtract offset from any increase in binary region size over source:
-    page_xywh['x'] -= 0.5 * max(0, page_image.width  - page_xywh['w'])
-    page_xywh['y'] -= 0.5 * max(0, page_image.height - page_xywh['h'])
+    page_xywh['x'] -= round(0.5 * max(0, page_image.width  - page_xywh['w']))
```
wrznr: See above.

ocrd_tesserocr/common.py
```diff
@@ -128,17 +261,28 @@ def image_from_region(workspace, region,
...
         # an ImageRegion which is properly contained in our TextRegion should
         # be completely masked, while partial overlap may be more difficult
-        # to decide (use polygons?)
+        # to decide. On the other hand, on the binary image, we can use
+        # connected component analysis to mask foreground areas which originate
+        # in the neighbouring regions. But that would introduce either the
+        # assumption that the input has already been binarized, or a dependency
+        # on some ad-hoc binarization method. Thus, it is preferable to use
+        # a dedicated processor for this (which produces clipped AlternativeImage
+        # or reduced polygon coordinates).
```
wrznr: To be discussed live.

ocrd_tesserocr/common.py
bertsky: Yes, the median of binarized images is almost always white. In fact, Ocropy uses this to determine whether the input image was inverted (by counting the pixels above vs below median). The only case where it does not work would be very tiny regions (say a page number, cropped too tightly).

bertsky: @wrznr just had another idea: an option would be to use `RGBA` mode (with transparency) in PIL objects throughout...

ocrd_tesserocr/common.py
bertsky: Technically, `BorderType` needs to have a `Coords` (`minoccurs=1`), and `CoordsType` needs to have a `@points` (`required=True`). But the input PAGE could be invalid, and then yes, this would lead to entering line 85 without `page_polygon`... I will fix.

bertsky: No I won't. If the PAGE is invalid, an exception is inevitable anyway.

ocrd_tesserocr/common.py
bertsky: see below under discussion of transparency instead of median

ocrd_tesserocr/common.py
bertsky: See [above](https://github.com/OCR-D/ocrd_tesserocr/pull/56#issuecomment-511425213): I could rewrite so page orientation gets active automatically as soon as the new schema version is merged into core.

ocrd_tesserocr/common.py
kba: It is rotated around the middle point, hence the `0.5 * width/height`, right?

ocrd_tesserocr/common.py
kba: To get @orientation we need PAGE 2019. But as we discussed yesterday, I will wait with that until @bertsky has time to formulate his idea on consistent URL for PAGE with XML version mechanism. Until then, use 2018 throughout, e.g. https://github.com/OCR-D/format-converters/pull/5

ocrd_tesserocr/common.py
bertsky: Exactly. The center of rotation for the image (which we have to compensate for when rotating the relative coordinates as well) was half-way into the image.

ocrd_tesserocr/common.py
bertsky: This is another `0.5`, this time it compensates for the _increase_ in image size due to rotation. This increase manifests equally left and right, above and below the image. Hence, the offset correction needs _half_ the difference in size.

ocrd_tesserocr/common.py
bertsky: Resolved alright? Feel free to post our result here (I don't quite remember).

bertsky:
Ok. So I will anticipate that discussion by changing the API to just 2 functions (without any redundancy in the implementation) – in a **new PR** (which we can easily abandon if we see reasons to). I will merge this one now.

ocrd_tesserocr/common.py
kba: https://github.com/OCR-D/assets/issues/34

> We will replace PrintSpace with Border and in the spec (ocr-d.github.io/gt//trans_documentation/lySatzspiegel.html) we add information for using Border.

@tboenig Let me know if you need help with automating the conversion.

PR #57, opened 2019-07-17
segment-region: use BlockPolygon

bertsky:
Now that sirfz/tesserocr#185 is merged, we can annotate Tesseract's polygon results instead of just the bounding box. Still not making `crop_polygons` true by default, because I believe these things cannot be trusted. 

I have linked with `tesserocr` requirement from Github instead of the PyPI pkg, because a new release will likely not arrive before Tesseract 5.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 ocrd_tesserocr
 ==============
 
-    Segment region, line, recognize with tesserocr
+    Segment into regions or lines, deskew, or recognize with tesserocr
 
 .. image:: https://circleci.com/gh/OCR-D/ocrd_tesserocr.svg?style=svg
     :target: https://circleci.com/gh/OCR-D/ocrd_tesserocr
@@ -12,6 +12,17 @@ ocrd_tesserocr
 .. image:: https://img.shields.io/docker/automated/ocrd/tesserocr.svg
     :target: https://hub.docker.com/r/ocrd/tesserocr/tags/
     :alt: Docker Automated build
+
+Introduction
+------------
+
+This offers `OCR-D`_ compliant workspace processors for (much of) the functionality of `Tesseract`_ via its Python API wrapper `tesserocr`_ . (Each processor is a step in the OCR-D functional model, and can be replaced with an alternative implementation. Data is represented within METS/PAGE.)
+
+This includes image preprocessing (cropping, binarization, deskewing), layout analysis (region, line, word segmentation) and OCR proper. Most processors can operate on different levels of the PAGE hierarchy, depending on the workflow configuration. Image results are referenced (read and written) via ``AlternativeImage``, text results via ``TextEquiv``.
+
+.. _OCR-D: https://ocr-d.github.io
+.. _Tesseract: https://github.com/tesseract-ocr
+.. _tesserocr: https://github.com/sirfz/tesserocr
 
 
 Installation
@@ -25,8 +36,9 @@ Required ubuntu packages:
 
 ::
 
-    pip install -r requirements
-    pip install .
+    make deps-ubuntu # or manually
+    make deps # or pip install -r requirements
+    make install # or pip install .
 
 If tesserocr fails to compile with an error:::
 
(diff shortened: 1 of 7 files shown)


bertsky:
Sorry about the force-pushing noise: having a hard time getting used to RST.

kba:
You can also use Markdown for the README if you prefer that, just have to change the setup.py so it's correctly displayed on PyPI.

requirements.txt
```diff
@@ -1,3 +1,3 @@
-ocrd >= 1.0.0b5
+ocrd >= 1.0.0b10
 click
-tesserocr >= 2.3.1
+tesserocr@git+https://github.com/sirfz/tesserocr#egg=tesserocr-2.4.0
```
kba: I could update https://pypi.org/project/ocrd-fork-tesserocr/ with a build of the current master of tesserocr. Using git links as dependency is somewhat brittle in my experience.

bertsky: Yes, I just noticed that `pip` does not always consider such releases as "newest", so it would not update over an existing 2.4.0 installation.

Using your PyPI release appears better to me, too. If you publish (under some version number as you like), I will change `requirements.txt` here accordingly.

README.rst
```diff
@@ -1,7 +1,7 @@
 ocrd_tesserocr
 ==============
 
-    Segment region, line, recognize with tesserocr
+    Segment into regions or lines, deskew, or recognize with tesserocr
```
wrznr: Cropping? (Even binarization?)

README.rst
```diff
@@ -13,6 +13,17 @@ ocrd_tesserocr
     :target: https://hub.docker.com/r/ocrd/tesserocr/tags/
     :alt: Docker Automated build
 
+Introduction
+------------
+
+This offers `OCR-D`_ compliant workspace processors for (much of) the functionality of `Tesseract`_ via its Python API wrapper `tesserocr`_ . (Each processor is a step in the OCR-D functional model, and can be replaced with an alternative implementation. Data is represented within METS/PAGE.)
+
+This includes image preprocessing (cropping, binarization, deskewing), layout analysis (region, line, word segmentation) and OCR proper. Most processors can operate on different levels of the PAGE hierarchy, depending on the workflow configuration. Image results are referenced (read and written) via ``AlternativeImage``, text results via ``TextEquiv``.
```
wrznr: Border, Orientation?

ocrd_tesserocr/crop.py
```diff
@@ -125,12 +125,12 @@ def process(self):
                     bin_bbox = image.getbbox()
                     if not bin_bbox:
                         # this does happen!
-                        LOG.debug("Ignoring region '%s' because its binarization is empty", ID)
+                        LOG.info("Ignoring region '%s' because its binarization is empty", ID)
```
wrznr: So, you decided to take the *more-verbose* road. We have to see if this is feasible in the long run.

ocrd_tesserocr/segment_region.py
```diff
@@ -171,18 +172,19 @@ def _process_page(self, it, page, page_image, page_xywh, page_id, file_id):
...
-            #     # avoid negative results (invalid in PAGE):
-            #     polygon = [(max(0, x + page_xywh['x']),
-            #                 max(0, y + page_xywh['y']))
-            #                for x, y in polygon]
-            #     points = points_from_polygon(polygon)
+            # sometimes these polygons are not planar, which causes
+            # PIL.ImageDraw.Draw.polygon (and likely others as well)
+            # to misbehave; unfortunately, we do not have coordinate
+            # semantics in PAGE (left/right inner/outer, multi-path etc)
+            # (probably a bug in Tesseract itself):
```
wrznr: Is there a way to check for such problematic cases (and do something about it)?

requirements.txt
kba: Published https://pypi.org/project/ocrd-fork-tesserocr/3.0.0rc2/ which includes https://github.com/sirfz/tesserocr/pull/185. It's only a source build, cannot get a wheel built but it should still work, albeit with compilation when pip installing it.

ocrd_tesserocr/segment_region.py
bertsky: This will be one of the responsibilities of OCR-D/core#252. See also PRImA-Research-Lab/PAGE-XML#13.

ocrd_tesserocr/crop.py
bertsky: Sorry, then I misunderstood you. So you want this to stay in DEBUG here, and downgrade similar statements in `segment-region` to DEBUG as well?

README.rst
bertsky: done.

README.rst
bertsky: done.

requirements.txt
bertsky: done.

bertsky:
Sorry, I should have made the last 3 on a separate branch/PR. Just sneaking by your approval!

kba:
Sneaky sneaky 😁

PR #58, opened 2019-07-18
simplify common:

bertsky:
- move resolving @imageFilename into `image_from_page`, too
- move OcrdExif there as well (cannot be done outside
  anymore, because AlternativeImage is not as reliable)
- unify image_from_* for all segments below page into
  `image_from_segment`, check whether rotation/orientation
  is present in segment/parent dynamically

### DIFF
--- a/ocrd_tesserocr/binarize.py
+++ b/ocrd_tesserocr/binarize.py
@@ -23,8 +23,7 @@ from ocrd import Processor
 from .config import TESSDATA_PREFIX, OCRD_TOOL
 from .common import (
     image_from_page,
-    image_from_region,
-    image_from_line,
+    image_from_segment,
     save_image_file,
     membername
 )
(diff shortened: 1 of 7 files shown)


bertsky:
As discussed earlier, this is another proposal for a somewhat simpler API with less code duplication in the implementation.

bertsky:
@kba I am agnostic over which variant makes the race for core.

ocrd_tesserocr/binarize.py
```diff
@@ -23,8 +23,7 @@
 from .config import TESSDATA_PREFIX, OCRD_TOOL
 from .common import (
     image_from_page,
-    image_from_region,
-    image_from_line,
+    image_from_segment,
```
wrznr: I like it!

ocrd_tesserocr/binarize.py
```diff
@@ -72,16 +71,15 @@ def process(self):
                                                                          value=self.parameter[name])
                                                                for name in self.parameter.keys()])]))
                 page = pcgts.get_Page()
-                page_image = self.workspace.resolve_image_as_pil(page.imageFilename)
+                page_image, page_xywh, _ = image_from_page(
```
wrznr: That is another function which has to be moved to `core`, right?

wrznr: Exactly. But a function from the workspace module was used prior to this PR. Treat it as a reminder. No worries.

bertsky: Indeed, we would want adopters to _forget_ all about `workspace.resolve_image_as_pil` and use the 2 new functions instead. Maybe we can even _hide_ the less abstract functions completely?

kba: `workspace.resolve_image_as_pil` could be marked as deprecated and phased out in the coming releases.

> Maybe we can even hide the less abstract functions completely?

What do you mean with "hide"?

bertsky: Sounds good. Hide: that would have been renaming with leading underscore, but of course, we do not want existing implementations to break.

bertsky:
I will merge now and update the ocropy side accordingly.

PR #62, opened 2019-07-31
Add rotation angle to orientation attribute on page level

wrznr:
With PAGE XML 2019, we can add skew angle on page level
appropriately!

### DIFF
--- a/ocrd_tesserocr/common.py
+++ b/ocrd_tesserocr/common.py
@@ -204,7 +204,7 @@ def image_from_page(workspace, page, page_id):
                  page_xywh['y'],
                  page_xywh['x'] + page_xywh['w'],
                  page_xywh['y'] + page_xywh['h']))
-        if page_xywh['angle']:
+        if 'angle' in page_xywh and page_xywh['angle']:
             LOG.info("About to rotate page '%s' by %.2f°",
                       page_id, page_xywh['angle'])
             page_image = page_image.rotate(page_xywh['angle'],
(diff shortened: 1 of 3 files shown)


wrznr:
@n00blet FYI

ocrd_tesserocr/common.py
```diff
@@ -173,10 +173,9 @@ def image_from_page(workspace, page, page_id):
...
                  'w': page_image.width,
                  'h': page_image.height}
-    # FIXME: uncomment as soon as we get @orientation in PageType:
-    # # region angle: PAGE orientation is defined clockwise,
-    # # whereas PIL/ndimage rotation is in mathematical direction:
-    # page_xywh['angle'] = -(page.get_orientation() or 0)
+    # region angle: PAGE orientation is defined clockwise,
+    # whereas PIL/ndimage rotation is in mathematical direction:
+    page_xywh['angle'] = -(page.get_orientation() or 0)
     # FIXME: remove PrintSpace here as soon as GT abides by the PAGE standard:
```
kba: @tboenig @VolkerHartmann just a reminder to revisit once you're ready with the updated GT

bertsky:
BTW, I don't think I can do the same thing for `cis-ocrd.ocropy.deskew` and `cis-ocrd.ocropy.common` easily, because that would entail forcing the entire MP of CIS to 1.0.0b11. (Also, @finkf has not merged the existing PR yet.)

Issue #49, opened 2019-07-03
Use common functionalities from "ocrd-utils"


wrznr:
As soon as they are available, the common functionalities from `core` should be employed.

bertsky:
Right, we should move `common` into core, but the image functions all depend on the workspace instance, so I think `ocrd.workspace.Workspace` would be the best place for them. (The coordinate conversion functions should of course go to `ocrd_utils`.)

kba:
Most of the `ocrd_*` modules have their own `utils` modules that are too specific for `ocrd_utils`. 

For the `image_from_*` functions, it would indeed make most sense to have them in workspace as you proposed.

bertsky:
Splendid. Just note, I will add a PR in ocrd_tesserocr which will rebase everything in `common` on true polygons (and masking instead of cropping). We should wait for that version in core.

kba:
Has been implemented as of https://github.com/OCR-D/ocrd_tesserocr/pull/66

PR #66, opened 2019-08-01
Adapt to utils moved to core, #49

kba:
This is the companion PR to https://github.com/OCR-D/core/pull/268


ocrd_tesserocr/common.py
```diff
@@ -1,151 +1,24 @@
...
-    Return a numpy array of the resulting polygon.
-    """
-    angle = np.deg2rad(angle)
-    cos = np.cos(angle)
-    sin = np.sin(angle)
-    # active rotation:  [[cos, -sin], [sin, cos]]
-    # passive rotation: [[cos, sin], [-sin, cos]] (inverse)
-    return orig + np.dot(polygon - orig, np.array([[cos, sin], [-sin, cos]]).transpose())
 
 # to be refactored into core (as method of ocrd.workspace.Workspace):
```
bertsky: Comment can be removed, too.

bertsky:
Strange – my 2 comments did not end up in the context of the diffs I placed them. I will hide them and re-review.

kba:
> Really? (I thought that seeing those FIXMEs in the report one by one is a major reason for running pylint?)

Was an oversight, just disabled to concentrate on code changes, will fix.

> Could you delete the file altogether (to avoid confusion in the future)?

Which file do you mean? pylintrc?

bertsky:
> > Could you delete the file altogether (to avoid confusion in the future)?
> 
> Which file do you mean? pylintrc?

No, `common.py`

PR #72, opened 2019-08-25
Catch exception when setting the locale

stweil:
Add also a comment that setting the locale is no longer needed
with newer versions of Tesseract.

### DIFF
--- a/ocrd_tesserocr/__init__.py
+++ b/ocrd_tesserocr/__init__.py
@@ -1,10 +1,18 @@
 import locale
-# circumvent tesseract-ocr issue #1670
+# Circumvent tesseract-ocr issue #1670
 # (which cannot be done on command line
 # because Click requires an UTF-8 locale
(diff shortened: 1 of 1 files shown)


kba:
Looks good, thanks. Instead of silently ignoring exception, version check would probably more robust. But if we document that we expect `tesseract >= 4.1.0` to be installed in the README.md (which is woefully out-of-date...), the problem would solve itself soon.

bertsky:
I concur – this is helpful already and should be merged independently. 

Before we can remove the workaround and point to 4.1.0 as dependency, we must increase test coverage (especially with `BlockPolygon`) and verify the precise requirements.

stweil:
I tried to use the version check, but with the current `tesserocr` the version can only be checked after the Tesseract initialization which triggers the assertion. :-(

That needs a fix in the `tesserocr` code.

Issue #61, opened 2019-07-19
prefer second position of output_file_grp for AlternativeImage, make FILEGRP_IMG fallback


bertsky:
Fixed by #59 for binarization already, this needs to be done for all processors. Moreover, this must be reflected in the documentation (`ocrd-tool.json`, README, docstrings) and tests.

bertsky:
Addition: since core v2.0.2 (process validation of file groups and parameters) this becomes a must, because without proper description in the tool json we now get:
```
Exception: fileGrp must not contain commas
```

bertsky:
Note: this issue has long since become **only** about documentation and automatic tests. Probably depends on a better spec.

bertsky:
Since https://github.com/OCR-D/spec/pull/164 this became obsolete.

PR #75, opened 2019-08-28
move to AlternativeImage feature selectors in OCR-D/core#294:

bertsky:
- all: use second output position as fileGrp USE to produce
  AlternativeImage
- all: rid of MetadataItem/Labels-related FIXME:
  with the updated PAGE model, we can now use
  @externalModel and @externalId
- all: use OcrdExif.resolution instead of xResolution
- all: create images with monotonically growing @comments
  (features)
- crop: use ocrd_utils.crop_image instead of PIL.Image.crop
- crop: fix bug when trying to access page_image if there
  are already region coordinates that we are ignoring
- crop: filter images already deskewed and cropped!
  (we must crop ourselves, and deskewing can not happen
   until afterwards)
- deskew: fix bugs in configuration-dependent corner cases
  related to whether deskewing has already been applied
  (on the page or region level):
  - for the page image, never use images already rotated
    (both for page level and region level processing,
     but for the region level, do rotate images ad hoc
     if @orientation is present on the page level)
  - for the region image, never use images already rotated
    (except for our own page-level rotation)
- segment-region: forgot to add feature "cropped" when
  producing cropped images

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+  * Adapt to feature selection/filtering mechanism for derived images in core
+  * Fixes for image feature-related corner cases in crop and deskew
+  * Use explicit (second) output fileGrp when producing derived images
+
 ## [0.4.0] - 2019-08-21
 
 Changed:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
   * Adapt to feature selection/filtering mechanism for derived images in core
-  * Fixes for image feature-related corner cases in crop and deskew
+  * Fixes for image-feature-related corner cases in crop and deskew
   * Use explicit (second) output fileGrp when producing derived images
 
 ## [0.4.0] - 2019-08-21
--- a/ocrd_tesserocr/binarize.py
+++ b/ocrd_tesserocr/binarize.py
@@ -49,7 +49,7 @@ class TesserocrBinarize(Processor):
         Set up Tesseract to recognize the segment image's layout, and get
         the binarized image. Create an image file, and reference it as
         AlternativeImage in the segment element. Add the new image file
-        to the workspace with with the fileGrp USE given in the second position
+        to the workspace with the fileGrp USE given in the second position
         of the output fileGrp, or ``OCR-D-IMG-BIN``, and an ID based on input
         file and input element.
         
--- a/ocrd_tesserocr/binarize.py
+++ b/ocrd_tesserocr/binarize.py
@@ -32,6 +32,13 @@ class TesserocrBinarize(Processor):
         kwargs['ocrd_tool'] = OCRD_TOOL['tools'][TOOL]
         kwargs['version'] = OCRD_TOOL['version']
         super(TesserocrBinarize, self).__init__(*args, **kwargs)
+        if hasattr(self, 'output_file_grp'):
+            try:
+                self.page_grp, self.image_grp = self.output_file_grp.split(',')
+            except ValueError:
+                self.page_grp = self.output_file_grp
+                self.image_grp = FALLBACK_IMAGE_GRP
+                LOG.info("No output file group for images specified, falling back to '%s'", FALLBACK_IMAGE_GRP)
 
     def process(self):
         """Performs binarization of the region / line with Tesseract on the workspace.
@@ -41,37 +48,36 @@ class TesserocrBinarize(Processor):
         
         Set up Tesseract to recognize the segment image's layout, and get
         the binarized image. Create an image file, and reference it as
-        AlternativeImage in the element and as file with a fileGrp USE
-        equal `OCR-D-IMG-BIN` in the workspace.
+        AlternativeImage in the segment element. Add the new image file
+        to the workspace with with the fileGrp USE given in the second position
+        of the output fileGrp, or ``OCR-D-IMG-BIN``, and an ID based on input
+        file and input element.
         
         Produce a new output file by serialising the resulting hierarchy.
         """
-        # pylint: disable=attribute-defined-outside-init
-        try:
-            self.page_grp, self.image_grp = self.output_file_grp.split(',')
-        except ValueError:
-            self.page_grp = self.output_file_grp
-            self.image_grp = FALLBACK_IMAGE_GRP
-            LOG.info("No output file group for images specified, falling back to '%s'", FALLBACK_IMAGE_GRP)
         oplevel = self.parameter['operation_level']
+        
         with PyTessBaseAPI(path=TESSDATA_PREFIX) as tessapi:
             for n, input_file in enumerate(self.input_files):
                 file_id = input_file.ID.replace(self.input_file_grp, self.image_grp)
                 page_id = input_file.pageId or input_file.ID
                 LOG.info("INPUT FILE %i / %s", n, page_id)
                 pcgts = page_from_file(self.workspace.download_file(input_file))
+                page = pcgts.get_Page()
+                
+                # add metadata about this operation and its runtime parameters:
                 metadata = pcgts.get_Metadata() # ensured by from_file()
                 metadata.add_MetadataItem(
                     MetadataItemType(type_="processingStep",
                                      name=self.ocrd_tool['steps'][0],
                                      value=TOOL,
-                                     # FIXME: externalRef is invalid by pagecontent.xsd, but ocrd does not reflect this
-                                     # what we want here is `externalModel="ocrd-tool" externalId="parameters"`
-                                     Labels=[LabelsType(#externalRef="parameters",
-                                                        Label=[LabelType(type_=name,
-                                                                         value=self.parameter[name])
-                                                               for name in self.parameter.keys()])]))
-                page = pcgts.get_Page()
+                                     Labels=[LabelsType(
+                                         externalModel="ocrd-tool",
+                                         externalId="parameters",
+                                         Label=[LabelType(type_=name,
+                                                          value=self.parameter[name])
+                                                for name in self.parameter.keys()])]))
+                
                 page_image, page_xywh, _ = self.workspace.image_from_page(
                     page, page_id)
                 LOG.info("Binarizing on '%s' level in page '%s'", oplevel, page_id)
@@ -129,5 +135,6 @@ class TesserocrBinarize(Processor):
                                     page_id=page_id,
                                     file_grp=self.image_grp)
         # update PAGE (reference the image file):
+        features = xywh['features'] + ",binarized"
         segment.add_AlternativeImage(AlternativeImageType(
-            filename=file_path, comments="binarized"))
+            filename=file_path, comments=features))
--- a/ocrd_tesserocr/crop.py
+++ b/ocrd_tesserocr/crop.py
@@ -4,6 +4,7 @@ import os.path
 import tesserocr
 from ocrd_utils import (
     getLogger, concat_padded,
+    crop_image,
     bbox_from_points, points_from_bbox, bbox_from_xywh,
     MIMETYPE_PAGE
 )
@@ -22,7 +23,7 @@ from .config import TESSDATA_PREFIX, OCRD_TOOL
 
 TOOL = 'ocrd-tesserocr-crop'
 LOG = getLogger('processor.TesserocrCrop')
-FILEGRP_IMG = 'OCR-D-IMG-CROP'
+FALLBACK_FILEGRP_IMG = 'OCR-D-IMG-CROP'
 
 class TesserocrCrop(Processor):
 
@@ -30,6 +31,14 @@ class TesserocrCrop(Processor):
         kwargs['ocrd_tool'] = OCRD_TOOL['tools'][TOOL]
         kwargs['version'] = OCRD_TOOL['version']
         super(TesserocrCrop, self).__init__(*args, **kwargs)
+        if hasattr(self, 'output_file_grp'):
+            try:
+                self.page_grp, self.image_grp = self.output_file_grp.split(',')
+            except ValueError:
+                self.page_grp = self.output_file_grp
+                self.image_grp = FALLBACK_FILEGRP_IMG
+                LOG.info("No output file group for images specified, falling back to '%s'",
+                         FALLBACK_FILEGRP_IMG)
 
     def process(self):
         """Performs page cropping with Tesseract on the workspace.
@@ -38,6 +47,12 @@ class TesserocrCrop(Processor):
         Set up Tesseract to detect text blocks on each page, and find
         the largest coordinate extent spanning all of them. Use this
         extent in defining a Border, and add that to the page.
+        
+        Moreover, crop the original image accordingly, and reference the
+        resulting image file as AlternativeImage in the Page element.
+        Add the new image file to the workspace with the fileGrp USE given
+        in the second position of the output fileGrp, or ``OCR-D-IMG-CROP``,
+        and an ID based on input file and input element.
         
         Produce new output files by serialising the resulting hierarchy.
         """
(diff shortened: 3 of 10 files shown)


bertsky:
Also fixes #61.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+  * Adapt to feature selection/filtering mechanism for derived images in core
+  * Fixes for image feature-related corner cases in crop and deskew
```
wrznr: ⛏ y: *image feature-related* -> *image-feature-related* (it is an adjective compound, https://www.grammarbook.com/punctuation/hyphens.asp)

ocrd_tesserocr/binarize.py
```diff
@@ -32,6 +32,13 @@ def __init__(self, *args, **kwargs):
         kwargs['ocrd_tool'] = OCRD_TOOL['tools'][TOOL]
         kwargs['version'] = OCRD_TOOL['version']
         super(TesserocrBinarize, self).__init__(*args, **kwargs)
+        if hasattr(self, 'output_file_grp'):
+            try:
+                self.page_grp, self.image_grp = self.output_file_grp.split(',')
```
wrznr: `split` results in a list, right? Not sure about the assignment here. What if we have only one output file group or three?

wrznr: Ah, both cases may result in `ValueError`. But the exception does not cover `A,B,C`.

ocrd_tesserocr/binarize.py
```diff
@@ -41,37 +48,36 @@ def process(self):
         
         Set up Tesseract to recognize the segment image's layout, and get
         the binarized image. Create an image file, and reference it as
-        AlternativeImage in the element and as file with a fileGrp USE
-        equal `OCR-D-IMG-BIN` in the workspace.
+        AlternativeImage in the segment element. Add the new image file
+        to the workspace with with the fileGrp USE given in the second position
```
wrznr: Double `with`.

ocrd_tesserocr/binarize.py
```diff
@@ -41,37 +48,36 @@ def process(self):
...
+        of the output fileGrp, or ``OCR-D-IMG-BIN``, and an ID based on input
```
wrznr: Why the double back ticks?

ocrd_tesserocr/crop.py
```diff
@@ -22,14 +23,22 @@
...
 
 class TesserocrCrop(Processor):
 
     def __init__(self, *args, **kwargs):
         kwargs['ocrd_tool'] = OCRD_TOOL['tools'][TOOL]
         kwargs['version'] = OCRD_TOOL['version']
         super(TesserocrCrop, self).__init__(*args, **kwargs)
+        if hasattr(self, 'output_file_grp'):
+            try:
+                self.page_grp, self.image_grp = self.output_file_grp.split(',')
```
wrznr: Cf. above.

ocrd_tesserocr/crop.py
```diff
@@ -39,6 +48,12 @@ def process(self):
         the largest coordinate extent spanning all of them. Use this
         extent in defining a Border, and add that to the page.
         
+        Moreover, crop the original image accordingly, and reference the
+        resulting image file as AlternativeImage in the Page element.
+        Add the new image file to the workspace with the fileGrp USE given
+        in the second position of the output fileGrp, or ``OCR-D-IMG-CROP``,
```
wrznr: See above.

ocrd_tesserocr/crop.py
```diff
@@ -53,27 +68,47 @@ def process(self):
...
-                                                               for name in self.parameter.keys()])]))
-                page = pcgts.get_Page()
+                                     Labels=[LabelsType(
+                                         externalModel="ocrd-tool",
+                                         externalId="parameters",
+                                         Label=[LabelType(type_=name,
+                                                          value=self.parameter[name])
+                                                for name in self.parameter.keys()])]))
+                
+                # warn of existing Border:
```
wrznr: *warn about an existing Border*

CHANGELOG.md
bertsky: Inclined to disagree: English does not feature _Durchkopplung_ hyphenation, and `image feature` itself is a non-hyphenated compound noun. But here it would help prevent a garden path effect, so I should either hyphenate all or none.

ocrd_tesserocr/binarize.py
bertsky: What do you mean the exception does not cover `A,B,C`? If anything except two output file groups are given, we enter the catch clause, as it should be. I don't see the problem...

ocrd_tesserocr/binarize.py
bertsky: thx!

ocrd_tesserocr/binarize.py
bertsky: Because by [PEP 287](https://www.python.org/dev/peps/pep-0287), docstrings are to be formatted in reStructuredText, if anything. This is also expected by tools like Sphinx.

CHANGELOG.md
wrznr: Pls. check the reference I added!

> Rule 1. Generally, hyphenate two or more words when they come before a noun they modify and act as a single idea. This is called a compound adjective.

ocrd_tesserocr/binarize.py
wrznr: And when you enter the catch clause in the `A,B,C` case, you assign `self.page_grp` with `A,B,C`. How is that supposed to work?

ocrd_tesserocr/crop.py
bertsky: Both forms are correct.

ocrd_tesserocr/segment_region.py
```diff
@@ -63,9 +67,12 @@ def process(self):
...
         
         If `crop_polygons` is True, create a cropped (and possibly deskewed)
-        raw image file for each region (masked along its polygon outline),
-        and reference it as AlternativeImage in the region element and
-        as file with a fileGrp USE equal `OCR-D-IMG-CROP` in the workspace.
+        image (without extra binarization) for each region (which gets
+        clipped to white outside its polygon outline), and reference th
+        resulting image file as AlternativeImage in the region element.
+        Add the new image to the workspace with the fileGrp USE given
+        in the second position of the output fileGrp, or `OCR-D-IMG-CROP`,
```
wrznr: Use double back ticks here 😸 !

bertsky:
> Technically, the changes proposed here carry over to `OCRopus` and `anybaseocr`. Does it make sense to add abstract wrappers to `core` for the single processing steps (i.e. `ProcessorCrop`) from which the module project implementations could derive?

Good idea. This would prevent making the same errors elsewhere, and avoid copying code. But it would probably be difficult to encapsulate the various fixed parts of the `process` method, with custom code scattered across different conditionals and loop bodies.

CHANGELOG.md
bertsky: I did. But please bear in mind prescriptive grammar like this rarely accounts for the true semantics of the system, esp. in non-trivial cases. The formulation of _rule 1_ is somewhat imprecise (`single idea`), and I believe it is meant for set phrases (`state-of-the-art`, `matter-of-fact`) and multi-word lexemes (`dressed-up`, `burned-out`, `multi-word`) rather than recursive compounds. The only thing in that grammar which gets close is _rule 9a_, which states that proper nouns (`Academy Award`, `African American`, `Middle Eastern`) are _not_ to be hyphenated through.

ocrd_tesserocr/binarize.py
bertsky: Oh, I see. Well, it is not supposed to work any better than without having all the candy here. That is, at some point later on, this will fail.

But I agree with you all these errors should be reported properly and as early as possible. So either the `Processor` base class or `process_cli` / `ocrd_cli_wrap_processor` should check in `ocrd-tool` for the number of outputs. We [discussed this in the spec](https://github.com/OCR-D/spec/issues/117), but it is not official, and not implemented yet. Also related to https://github.com/OCR-D/core/pull/296.

ocrd_tesserocr/segment_region.py
bertsky: Sorry, forgot this one.

setup.py
```diff
@@ -16,7 +16,7 @@
 
 setup(
     name='ocrd_tesserocr',
-    version='0.4.0',
+    version='0.4.1',
```
kba: Better to do that in explicit version bumping commits.

kba:
> Good idea. This would prevent making the same errors elsewhere, and avoid copying code. But it would probably be difficult to encapsulate the various fixed parts of the process method, with custom code scattered across different conditionals and loop bodies.

Yeah, that would be tricky without changing the API and breaking existing code. It's a neat idea though and I wish we had a wrapper around the `process` method that could accomodate specialization and do startup/cleanup work but we haven't.

(e.g. requiring processors to implement a protected `_process` method that the base class' `process` method would call. But it's too late for that now)

ocrd_tesserocr/binarize.py
kba: It could be cleaner and will be, these issues will be adressed but for now, I don't think this is a showstopper.

setup.py
bertsky: Understood. @kba, should I revert this in the PR? Do you want to do the merge and the versioning commit along with the new release?

bertsky:
The last 2 commits explained:

I noticed that Tesseract internally uses 8 bit grayscale as feed for the LSTM models. So I gathered its binarization is only needed for the non-LSTM models and layout analysis, and therefore added `feature_filter='binarized'` (i.e. requested raw images from the API). But then it turned out that existing models (in fact, existing training procedures) _also_ feed binarized data. So the network would be quite perplexed to see grayscale input at runtime. This is also what I could validate by measurements on GT (tentatively): CER shoots up when using raw images for `SetImage`. And it shoots up even more when cropped/rotated/masked areas do not get filled with background colour but with white or transparent (since transparency is reduced to white internally). So the network is more confused about white background re-appearing in a grayscale image than white disappearing completely at runtime.

Anyway, the revert is necessary to keep up the expectations of _current models_, but the original commit could be re-activated when we have a different training procedure!

bertsky:
Here are my CER measurements (in percent) on 2 GT bags with textual annotation, in a workflow configuration similar to [this](https://gist.github.com/bertsky/18f836c941698832a60f6a3dd4cbf0e4) (i.e. including Olena _Wolf_ or Ocropy _nlbin-nrm_ binarization, Ocropy page-level deskewing, clipping and resegmentation, and dewarping).

| *Tesseract model* | *source* | *trained on* |
| ------- | ------- | ------- |
| Fraktur | [tessdata_best](https://github.com/tesseract-ocr/tessdata_best/) (Google) | binarized artificial images |
| GT4HistOCR | [tesstrain](https://github.com/OCR-D/ocrd-train/wiki/GT4HistOCR) (UB Mannheim) | greyscale-normalized scanned images |

On `euler_rechenkunst01_1738` (a bag with JPEG artifacts and lower resolution):

 | *input* | *Fraktur* | *GT4HistOCR* |
 | ------- | -------- | ------- |
 | binarized (Wolf) | 9.7 | 11.9 |
 | raw with transparent/white background | 19.7 | 9.9 |
 | raw with estimated background color | 12.9 | 61.3 |
 | greyscale-normalized (ocropy-nlbin) | 9.4 | 9.8 |

Results are similar in tendency for `weigel_gnothi02_1618` (which is cleaner):

 | *input* | *Fraktur* | *GT4HistOCR* |
 | ------- | -------- | ------- |
 | binarized (Wolf) | 14.0 | 7.3 |
 | raw with transparent/white background | 36.5 | 8.3 |
 | raw with estimated background color | 13.8 | 58.6 |
 | greyscale-normalized (ocropy-nlbin) | 13.5 | 7.0 |

So Tesseract gets perplexed …
- on a standard model (trained `binarized`):
    - … not at all by greyscale normalized line images,
    - … somewhat by median background in unnormalized greyscale line images,
    - … extremely by white background in unnormalized greyscale line images,
- on GT4HistOCR model (trained `grayscale_normalized`):
    - … somewhat by binarized line images,
    - … a little by white background in unnormalized greyscale line images,
    - … extremely by median background in unnormalized greyscale line images

wrznr:
It is striking that both, the stock and the GT4HistOCR model, perform so poorly! CER between 7 and 9 % is a) simply not good enough b) way below the numbers @stweil reported at OCR-D developer workshop.

stweil:
@wrznr, CER often needs a closer look, especially what kind of errors did occur (maybe missing normalization?).

@bertsky, how can I reproduce your setting?

bertsky:
True! That was also one of my [messages](https://hackmd.io/@FKFH0M1sR2SdJZwK5U8Cfg/S1YQ4NeNr#/5/37) at the workshop. Generally, I am quite certain this is due to the relatively bad quality of our GT: 
- very imprecise segmentation (often too coarse with much overlap, sometimes too tight with cutoff) and
- suboptimal image quality (often noisy/skewed/warped, sometimes compression artifacts). 

Despite all the preprocessing and resegmentation efforts, we are not able to squeeze less than 11% CER out of the whole dataset with the stock models. And I don't believe you would get much better results if you trained/finetuned a dedicated OCR model on our GT. But maybe @stweil wants to disprove that?

GT4HistOCR also _looks_ much cleaner. If really all they did for preprocessing was running `ocropus-nlbin` and `ocropus-gpageseg`, then their raw data already were a lot easier to begin with. (And as already argued, this might be simply the result of filtering out the lesser-quality lines.)

So, I think the situation demands:

- better **despeckling**: At the moment, we only have _ocrd-cis-ocropy-denoise_ (which wraps `ocrolib.remove_noise` for black noise, so maybe we at least should make it work for white noise as well: @Doreenruirui, I think I might just do that). But really it's time we finally got hold of some data-driven despeckling: @n00blet @mjenckel, can you offer something?

- better **deskewing**: We currently rely on _ocrd-cis-ocropy-deskew_ and _ocrd-tesserocr-deskew_, both of which improve overall results a little, but often enough fail miserably for no good (visually convincing) reason. Again, it's time we finally got some data-driven deskewing: I was [struggling](https://github.com/NVlabs/ocrorot/issues/3) to get [ocrorot](https://github.com/NVlabs/ocrorot) running, but I am sure there must be more tools out there. @n00blet @mjenckel, can you offer something?

- even better **binarization**

- semi-interactively improving GT segmentation (aided by our existing clipping/resegmentation/repair tools)

bertsky:
Example from our preprocessing/resegmentation pipeline: cleanly separated from neighbours, but still quite noisy

OCR-D-IMG-DEWARP_0001_r_3_1_tl_10

Example from GT4HistOCR: very clean, but skewed/warped

00008 nrm

wrznr:
@bertsky @stweil So our impression

> And as already argued, this might be simply the result of filtering out the lesser-quality lines.

that GT4HistOCR is suboptimal for OCR training is real.

> And I don't believe you would get much better results if you trained/finetuned a dedicated OCR model on our GT.

I am with you. **But** we could try to *noise* GT4HistOCR. Where could we further discuss this issue? (This PR is clearly not the right place.) A dedicated issue in `tesstrain`?

bertsky:
@stweil The [above link](https://gist.github.com/bertsky/18f836c941698832a60f6a3dd4cbf0e4) should be workable on the current versions. But it's probably best to use the PRs I have used: OCR-D/core#311 and OCR-D/ocrd_tesserocr#75 and cisocrgroup/ocrd_cis#16 and [master cor-asv-ann](https://github.com/ASVLeipzig/cor-asv-ann) (for evaluation).

CER measurement in cor-asv-ann-evaluate works as [documented](https://github.com/ASVLeipzig/cor-asv-ann#evaluation): vanilla Levenshtein, no normalization. This is best for comparability, results might look better (and be more fair) with different metrics and normalizations. The package offers some, but results look quite similar with other settings, even `NFC`.

bertsky:
> **But** we could try to _noise_ GT4HistOCR. Where could we further discuss this issue? (This PR is clearly not the right place.) A dedicated issue in `tesstrain`?

I agree, both degradation and binarization should be employed to make GT4HistOCR models robust.

wrznr:
@bertsky Is there PR in core which blocks merging here?

bertsky:
> Where could we further discuss this issue? (This PR is clearly not the right place.) A dedicated issue in `tesstrain`?

@stweil Is tesseract-ocr/tesstrain#73 the right place for this? Or better open a new issue strictly about binarization/augmentation (not specific to GT4HistOCR)?

BTW, according to my measurements, `Fraktur` performs similar on GT4HistOCR data, although not quite as bad as our GT. But of course, it really depends on the subcorpus/century:
- `dta19` and `RefCorpus-ENHG-Incunabula` and `Kallimachos`: 7.0% CER
- `dta19` only: 4.9% CER
Since our GT is more diverse / has less 19th century share, worse results can be expected (independent of bad segmentation / preprocessing).

bertsky:
>  Is there PR in core which blocks merging here?

No, not really. OCR-D/core#311 is related, but not a dependency. Thanks, I will merge this for now.

PR #79, opened 2019-10-14
Set locale to C instead of C.UTF-8

### LINKED ISSUE #78: locale setting for tesserocr does not work on ubuntu 19.04
Here:
https://github.com/OCR-D/ocrd_tesserocr/blob/1ba854016e0651e93b94b7c060314a533bf0446d/ocrd_tesserocr/__init__.py#L7
the locale should be set to `C.UTF-8`, but this doesn't work because the locale `C.UTF-8` gets aliased to `en_US.UTF-8` (at least on my ubuntu 19.04 system, see below)

Afterwards the import here:
https://github.com/OCR-D/ocrd_tesserocr/blob/1ba854016e0651e93b94b7c060314a533bf0446d/ocrd_tesserocr/recognize.py#L4
fails with the good old:

`!strcmp(locale, "C"):Error:Assert failed:in file baseapi.cpp, line 209`

The alias comes from the file:
`/usr/share/X11/locale/locale.alias`
In line 63:
`C.UTF-8                                 en_US.UTF-8`

This is the default as can be seen here:

https://github.com/mirror/libX11/blob/87c77a1e6d7034536e9d25ce24a667ebf53486a7/nls/locale.alias.pre#L18

hnesk:
Fixes #78

### DIFF
--- a/ocrd_tesserocr/__init__.py
+++ b/ocrd_tesserocr/__init__.py
@@ -10,7 +10,7 @@ import locale
 # is an exception.
 # pylint: disable=wrong-import-position
 try:
-    locale.setlocale(locale.LC_ALL, 'C.UTF-8')
+    locale.setlocale(locale.LC_ALL, 'C')
 except locale.Error:
     pass
 


stweil:
The solution is to use Tesseract 4.1 or newer which does not need locale settings.

bertsky:
> The solution is to use Tesseract 4.1 or newer which does not need locale settings.

I think we could probably condition this workaround in ocrd_tesserocr on which version is actually installed. (That would entail parsing the output of `tesserocr.tesseract_version()` and accepting both `4.1` and `5.0` builds as not requiring to reset localization.)

But then again – apart from installation issues – I don't think there is a real downside of still keeping the workaround in there, isn't there?

stweil:
Sure, setting the locale does not harm, so the workaround can be kept in place.

`tesserocr.tesseract_version()` currently already triggers the assertion in `4.0`, so using that is not possible without setting the right locale.

bertsky:
@stweil Oh, I see. Thanks for clarifying!

@hnesk Can we close?

hnesk:
I dug a bit deeper here: https://github.com/OCR-D/ocrd_tesserocr/issues/78#issuecomment-541971751 . This could be solved by checking the return value of `locale.getlocale(locale.LC_CTYPE)` (should == None or != 'en') after setting the locale. But it's fine with me to keep it as it is. Nice to see tesseract 4.1 not depending on locale settings anymore!

PR #83, opened 2019-10-26
[WIP] Changes that require ocrd >= 2.0.0a1

kba:
Do not merge until 2.0.0 is stable.


Dockerfile
```diff
@@ -9,17 +9,16 @@ WORKDIR /build-ocrd
...
 RUN apt-get update && \
     apt-get -y install --no-install-recommends \
-    ca-certificates \
-    make \
-    git
-COPY Makefile .
-RUN make deps-ubuntu
-COPY ocrd_tesserocr ./ocrd_tesserocr
+    libtesseract-dev \
+    libleptonica-dev \
```
stweil: `libleptonica-dev` is redundant (it is already a dependency of `libtesseract-dev`).

Dockerfile
```diff
@@ -9,17 +9,16 @@ WORKDIR /build-ocrd
...
+    tesseract-ocr-eng \
```
stweil: `tesseract-ocr-eng` is redundant, too.

Dockerfile
```diff
@@ -9,17 +9,16 @@ WORKDIR /build-ocrd
...
+COPY Makefile .
 RUN apt-get update && \
     apt-get -y install --no-install-recommends \
-    ca-certificates \
-    make \
-    git
-COPY Makefile .
-RUN make deps-ubuntu
-COPY ocrd_tesserocr ./ocrd_tesserocr
+    libtesseract-dev \
```
stweil: Does this install 4.1 (current stable release)? If it still installs 4.0 (no longer supported, known bugs), that should be changed.

The `ENV` statements in lines 5 and 6 were needed for Tesseract 4.0. Tesseract 4.1 does not require them.

README.md
```diff
@@ -8,59 +7,83 @@ ocrd\_tesserocr
...
------
+```sh
+docker pull ocrd/tesserocr
+```
+
+### From git 
+
+This is the best option if you want to change the source code or install the latest, unpublished changes.
+
+We strongly recommend to use a [virtualenv](https://virtualenv.pypa.io/en/stable/userguide/).
```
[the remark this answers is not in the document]
kba: Good point, fixed.

Dockerfile
kba: Thanks, will remove.

Dockerfile
kba: True, thanks.

Dockerfile
kba: Yes, this will install [4.1.0-1](https://www.ubuntuupdates.org/package/core/eoan/universe/base/tesseract)

PR #90, opened 2019-11-05
Add section on installing tesseract from PPA for Ubuntu < 19.10 #89 No description available

kba:


kba:
As long as the change is technically correct, I'm fine with respecting the Tesseract branding/spelling. 😉 Will fix.

PR #94, opened 2019-11-15
docker: Switch to 18.04 + @AlexanderP PPA

kba:


kba:
Happy to adapt this further but merging now, so `latest` ocrd/tesserocr docker image is rebuilt.

PR #95, opened 2019-11-23
Don't download unsupported old Tesseract

stweil:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,7 @@ COPY ocrd_tesserocr ./ocrd_tesserocr
 COPY Makefile .
 RUN apt-get update && \
     apt-get install -y software-properties-common && \
-    add-apt-repository ppa:alex-p/tesseract-ocr && \
-    apt-get update && \
+    add-apt-repository -y ppa:alex-p/tesseract-ocr && \
     apt-get -y install --no-install-recommends \
     libtesseract-dev \
     tesseract-ocr \
--- a/Makefile
+++ b/Makefile
@@ -21,12 +21,10 @@ help:
 	@echo "    deps-ubuntu   Dependencies for deployment in an ubuntu/debian linux"
 	@echo "                  (lib*-dev merely for building tesserocr with pip)"
 	@echo "                  (tesseract-ocr: Ubuntu 18.04 now ships 4.0.0,"
-	@echo "                   but some beta, not the final release,"
-	@echo "                   on which tesserocr 2.4.0 depends;"
-	@echo "                   this downloads a tesseract build from git"
-	@echo "                   and installs it system-wide -"
-	@echo "                   intended for dockerfile and travis,"
-	@echo "                   not recommended for live systems!)"
+	@echo "                   which is unsupported. Add the tesseract-ocr PPA"
+	@echo "                   from Alexander Pozdnyakov which provides 4.1.0."
+	@echo "                   See https://launchpad.net/~alex-p/+archive/ubuntu/tesseract-ocr"
+	@echo "                   for details.)"
 	@echo "    deps          Install python deps via pip"
 	@echo "    deps-test     Install testing python deps via pip"
 	@echo "    install       Install"
@@ -46,13 +44,11 @@ help:
 
 # Dependencies for deployment in an ubuntu/debian linux
 # (lib*-dev merely for building tesserocr with pip)
-# (tesseract-ocr: Ubuntu 18.04 now ships 4.0.0,
-#  but some beta, not the final release,
-#  on which tesserocr 2.4.0 depends;
-#  this downloads a tesseract build from git
-#  and installs it system-wide -
-#  intended for dockerfile and travis,
-#  not recommended for live systems!)
+# (tesseract-ocr: Ubuntu 18.04 now ships 4.0.0
+#  which is unsupported. Add the tesseract-ocr PPA
+#  from Alexander Pozdnyakov which provides 4.1.0.
+#  See https://launchpad.net/~alex-p/+archive/ubuntu/tesseract-ocr
+#  for details.)
 deps-ubuntu:
 	apt-get install -y \
 		git \
@@ -63,8 +59,6 @@ deps-ubuntu:
 		tesseract-ocr-eng \
 		tesseract-ocr \
 		wget
-	wget -O - https://github.com/nijel/tesseract-ocr-build/releases/download/4.0.0-1/tesseract.tar.xz | tar -xJf -
-	cp -rt /usr .tesseract/*
 
 # Install python deps via pip
 deps:


Makefile
```diff
@@ -63,8 +59,6 @@ deps-ubuntu:
 		tesseract-ocr-eng \
 		tesseract-ocr \
 		wget
-	wget -O - https://github.com/nijel/tesseract-ocr-build/releases/download/4.0.0-1/tesseract.tar.xz | tar -xJf -
-	cp -rt /usr .tesseract/*
```
bertsky: It's not acceptable to just remove this step entirely here. (And Circle-CI configuration is not the only use-case.)
```suggestion
	apt-get install -y software-properties-common && add-apt-repository ppa:alex-p/tesseract-ocr
```

.circleci/config.yml
```diff
@@ -8,6 +8,7 @@ jobs:
     docker:
       - image: ubuntu:18.04
     steps:
+      - run: apt-get update ; apt-get install -y software-properties-common ; add-apt-repository ppa:alex-p/tesseract-ocr
```
bertsky: already covered by (updated) `deps-ubuntu` – see above.
```suggestion
```

stweil: Thanks, fixed now.

Dockerfile
```diff
@@ -12,8 +12,7 @@ COPY ocrd_tesserocr ./ocrd_tesserocr
 COPY Makefile .
 RUN apt-get update && \
     apt-get install -y software-properties-common && \
-    add-apt-repository ppa:alex-p/tesseract-ocr && \
-    apt-get update && \
+    add-apt-repository -y ppa:alex-p/tesseract-ocr && \
     apt-get -y install --no-install-recommends \
     libtesseract-dev \
     tesseract-ocr \
```
bertsky: We should now also delegate to `deps-ubuntu` in the Dockerfile, of course.

stweil: I am not convinced that this is the right solution. It would be wrong on hosts running very recent versions of Ubuntu or Debian and for people running their own version of Tesseract. How should those cases be handled?

For Docker and CircleCI the situation is different as we have a well defined environment. Perhaps the Makefile rule can be restricted to Ubuntu 18.04?

bertsky: > I am not convinced that this is the right solution. It would be wrong on hosts running very recent versions of Ubuntu or Debian and for people running their own version of Tesseract. How should those cases be handled?

I assume with `this` you refer to adding the PPA under `deps-ubuntu` in the Makefile.

Where else would you encapsulate system dependencies? This is precisely the convention that was introduced for OCR-D processors, and specifically for Ubuntu 18.04 to keep this manageable. ocrd_tesserocr should continue leading by good example.

As for running your own Tesseract build: `deps-ubuntu` does not prevent that – it does not even interfere (as long as you don't overwrite standard package directories).

bertsky:
> @stweil I am not sure I used the `suggestion` syntax correctly (so you might have to edit manually), but AFAICS you did not fix the Makefile yet...

Now after the force-push this PR simply breaks the existing Makefile (and ocrd_all), and the Circle-CI and Docker setups hold redundant knowledge – and even as separate PRs, despite belonging together.

Dockerfile
stweil: Adding a PPA is very special, and I really would feel better if that part were only done for Ubuntu 18.04.

In addition we could look for `/usr/local/include/tesseract` (or some other indicator that there is a local installation). If there is already such an installation, we should not add a 2nd installation as that can cause much confusion.

stweil:
I suggest to merge pull request #96 which fixes the broken rules for Docker first.

Pull request #97 still fails during CircleCI, so it is not ready to get merged.
Neither is the current pull request #95 ready to get merged as there is still ongoing discussion how things should be handled.

Dockerfile
bertsky: > Adding a PPA is very special, and I really would feel better if that part were only done for Ubuntu 18.04.

But I just said `deps-ubuntu` denotes Ubuntu 18.04 under the current spec. 

You don't have to use that target if you don't like it, but it must be available when system dependencies exist, and must work automatically.

> In addition we could look for `/usr/local/include/tesseract` (or some other indicator that there is a local installation). If there is already such an installation, we should not add a 2nd installation as that can cause much confusion.

As I just said, `deps-ubuntu` does not interfere with local installation.  If the user wants that in addition, let her (keep that up-to-date). __Not doing__ system installation because of such heuristics (version checks??) would cause confusion.

bertsky:
> I suggest to merge pull request #96 which fixes the broken rules for Docker first.

As I've [stated above](https://github.com/OCR-D/ocrd_tesserocr/pull/95#discussion_r350187170), I am against scattering copies of this recipe in different places. (The same goes for #97.)

> Pull request #97 still fails during CircleCI, so it is not ready to get merged.
> Neither is the current pull request #95 ready to get merged as there is still ongoing discussion how things should be handled.

The CI failure here shows that you cannot separate this from #97, and the CI failure in #97 shows that the PPA method still does not work because its version is still too old: it does not contain `GetBestLSTMSymbolChoices`.

Dockerfile
stweil: Now I am confused. You said that `deps-ubuntu` denotes Ubuntu 18.04, but that not doing system installation because of such heuristics (version checks??) would cause confusion?

To be more explicit: I see no problem with `deps-ubuntu` as long as the documentation clearly says what it is intended for (and in the case of `ocrd_tesserocr` when it should not be used) and as long as it is never called unexpectedly. It is ok when it is triggered by a user running `sudo make deps-ubuntu` or in the root context of a Docker build or CircleCI.

When I use `ocrd_all` and have installed Tesseract using `make install-tesseract`, I clearly don't want a 2nd installation.

stweil:
This PR is superseded by PR #97, so closing it.

Dockerfile
bertsky: > Now I am confused. You said that `deps-ubuntu` denotes Ubuntu 18.04, but that not doing system installation because of such heuristics (version checks??) would cause confusion?

Yes, _by convention_! Not doing system-wide installation because of some checks in `/usr/local` would make that rule hard to predict and use.

> To be more explicit: I see no problem with `deps-ubuntu` as long as the documentation clearly says what it is intended for (and in the case of `ocrd_tesserocr` when it should not be used) and as long as it is never called unexpectedly. It is ok when it is triggered by a user running `sudo make deps-ubuntu` or in the root context of a Docker build or CircleCI.

But then we are in perfect agreement!

> When I use `ocrd_all` and have installed Tesseract using `make install-tesseract`, I clearly don't want a 2nd installation.

Then this is an issue specific to ocrd_all, it is off topic here. (You'd still want ocrd_tesserocr to encapsulate all of Tesseract's system dependencies though, if not Tesseract itself. Otherwise you end up copying recipes again.)

PR #96, opened 2019-11-25
Update rules in Dockerfile

stweil:
- Only install required packages for software-properties-common
- Add missing option `-y` for add-apt-repository
- Remove unneeded update after add-apt-repository
- Replace build-essential by the required packages g++ and make
- Remove libtesseract-dev when it is no longer needed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,12 +10,9 @@ COPY requirements.txt .
 COPY requirements_test.txt .
 COPY ocrd_tesserocr ./ocrd_tesserocr
 COPY Makefile .
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends software-properties-common && \
-    add-apt-repository -y ppa:alex-p/tesseract-ocr && \
+RUN make deps-ubuntu && \
     apt-get install -y --no-install-recommends \
-    g++ libtesseract-dev make \
-    tesseract-ocr \
+    g++ make \
     tesseract-ocr-script-frak \
     tesseract-ocr-deu \
     && make deps install \
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY ocrd_tesserocr ./ocrd_tesserocr
 COPY Makefile .
 RUN make deps-ubuntu && \
     apt-get install -y --no-install-recommends \
-    g++ make \
+    g++ \
     tesseract-ocr-script-frak \
     tesseract-ocr-deu \
     && make deps install \
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,15 +11,13 @@ COPY requirements_test.txt .
 COPY ocrd_tesserocr ./ocrd_tesserocr
 COPY Makefile .
 RUN apt-get update && \
-    apt-get install -y software-properties-common && \
-    add-apt-repository ppa:alex-p/tesseract-ocr && \
-    apt-get update && \
-    apt-get -y install --no-install-recommends \
-    libtesseract-dev \
+    apt-get install -y --no-install-recommends software-properties-common && \
+    add-apt-repository -y ppa:alex-p/tesseract-ocr && \
+    apt-get install -y --no-install-recommends \
+    g++ libtesseract-dev make \
     tesseract-ocr \
     tesseract-ocr-script-frak \
     tesseract-ocr-deu \
-    build-essential \
     && make deps install \
     && rm -rf /build-ocrd \
-    && apt-get -y remove --auto-remove build-essential
+    && apt-get -y remove --auto-remove g++ libtesseract-dev make


stweil:
Using `make deps-ubuntu` can be done in a followup commit.

bertsky:
> Using `make deps-ubuntu` can be done in a followup commit.

@stweil, could you please add that follow-up commit? (Here or in #97.)

stweil:
Done. Please review.

Dockerfile
```diff
@@ -10,16 +10,11 @@ COPY requirements.txt .
...
-RUN apt-get update && \
-    apt-get install -y software-properties-common && \
-    add-apt-repository ppa:alex-p/tesseract-ocr && \
-    apt-get update && \
-    apt-get -y install --no-install-recommends \
-    libtesseract-dev \
-    tesseract-ocr \
+RUN make deps-ubuntu && \
+    apt-get install -y --no-install-recommends \
+    g++ make \
```
kba: make is in ocrd/core

Dockerfile
```diff
@@ -10,16 +10,11 @@ COPY requirements.txt .
...
-RUN apt-get update && \
-    apt-get install -y software-properties-common && \
-    add-apt-repository ppa:alex-p/tesseract-ocr && \
-    apt-get update && \
-    apt-get -y install --no-install-recommends \
-    libtesseract-dev \
-    tesseract-ocr \
+RUN make deps-ubuntu && \
+    apt-get install -y --no-install-recommends \
+    g++ \
```
bertsky: So `g++` is stil needed to build `tesserocr` right?

kba: Yes, but I'd have no quarrel adding it to ocrd/core. I don't see a reason why 100MB more or less would matter at this point if it makes implementer's life easier.

stweil: I was not sure and just tried it. Yes, it is still used, but obviously only for linking:

    x86_64-linux-gnu-g++ -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,relro -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 build/temp.linux-x86_64-3.7/tesserocr.o -llept -ltesseract -o build/lib.linux-x86_64-3.7/tesserocr.cpython-37m-x86_64-linux-gnu.so

So I am pretty sure this dependency could be dropped with some efforts.

bertsky: > Yes, but I'd have no quarrel adding it to ocrd/core. I don't see a reason why 100MB more or less would matter at this point if it makes implementer's life easier.

I concur. But still I think this dependency should be explicit here – and not just in the Dockerfile, but in the Makefile under `deps-ubuntu`.



> Yes, it is still used, but obviously only for linking:
> 
> ```
> x86_64-linux-gnu-g++ -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,relro -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 build/temp.linux-x86_64-3.7/tesserocr.o -llept -ltesseract -o build/lib.linux-x86_64-3.7/tesserocr.cpython-37m-x86_64-linux-gnu.so
> ```
> 
> So I am pretty sure this dependency could be dropped with some efforts.

Yes, but that would require adding assumptions on the way the linker can be called from setuptools (which in turn might be dependent on Python / setuptools versions). Better encapsulate everything and be more robust.

Dockerfile
```diff
@@ -10,16 +10,11 @@ COPY requirements.txt .
...
     tesseract-ocr-script-frak \
     tesseract-ocr-deu \
-    build-essential \
     && make deps install \
     && rm -rf /build-ocrd \
-    && apt-get -y remove --auto-remove build-essential
+    && apt-get -y remove --auto-remove g++ libtesseract-dev make
```
bertsky: Wait, why would you remove `libtesseract-dev`? That will break `tesserocr` and `ocrd-tesserocr`, won't it?

kba: Yes, removing the headers and library will break tesserocr at runtime i think.

stweil: It's already handled by `make deps-ubuntu`.

kba: I have tested it and it is not a problem at runtime, works fine.

bertsky: Ok great – thanks for testing!

(It looks like all runtime dependencies are in `libtesseract4` already.)

Dockerfile
bertsky: So we could remove `g++` here now, but it does not matter really.

stweil:
Please merge - I don't have the necessary rights.

PR #97, opened 2019-11-25
circleci: Switch to @AlexanderP PPA

stweil:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,8 +9,6 @@ jobs:
       - image: ubuntu:18.04
     steps:
       - run: apt-get update
-      - run: apt-get install -y --no-install-recommends software-properties-common
-      - run: add-apt-repository -y ppa:alex-p/tesseract-ocr
       - run: apt-get install -y --no-install-recommends make git curl
       - checkout
       - run: make deps-ubuntu deps-test deps install repo/assets
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,10 @@ jobs:
     docker:
       - image: ubuntu:18.04
     steps:
-      - run: apt-get update ; apt-get install -y make git curl
+      - run: apt-get update
+      - run: apt-get install -y --no-install-recommends software-properties-common
+      - run: add-apt-repository -y ppa:alex-p/tesseract-ocr
+      - run: apt-get install -y --no-install-recommends make git curl
       - checkout
       - run: make deps-ubuntu deps-test deps install repo/assets
       - run: make test-cli
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,8 +8,7 @@ jobs:
     docker:
       - image: ubuntu:18.04
     steps:
-      - run: apt-get update
-      - run: apt-get install -y --no-install-recommends make git curl
+      - run: apt-get update && apt-get install -y --no-install-recommends make git curl
       - checkout
       - run: make deps-ubuntu deps-test deps install repo/assets
       - run: make test-cli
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -8,7 +8,8 @@ jobs:
     docker:
       - image: ubuntu:18.04
     steps:
-      - run: apt-get update ; apt-get install -y make git curl
+      - run: apt-get update
+      - run: apt-get install -y --no-install-recommends make git curl
```
bertsky: According to [Docker-build documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) it is not recommended to have _update_ and _install_ in separate steps. Not sure under what circumstances that is a problem though.

stweil: I fixed that with my latest commit.

Dockerfile
```diff
@@ -11,15 +11,13 @@ COPY requirements_test.txt .
 COPY ocrd_tesserocr ./ocrd_tesserocr
 COPY Makefile .
 RUN apt-get update && \
-    apt-get install -y software-properties-common && \
-    add-apt-repository ppa:alex-p/tesseract-ocr && \
-    apt-get update && \
-    apt-get -y install --no-install-recommends \
-    libtesseract-dev \
+    apt-get install -y --no-install-recommends software-properties-common && \
```
kba: Would it make sense to have `software-properties-common` in `ocrd/core` Dockerfile?

Dockerfile
```diff
@@ -11,15 +11,13 @@ COPY requirements_test.txt .
...
+    add-apt-repository -y ppa:alex-p/tesseract-ocr && \
+    apt-get install -y --no-install-recommends \
+    g++ libtesseract-dev make \
```
kba: `make` is provided by `ocrd/core`, no need to install it, you can uninstall it but it will make little difference I think.

Dockerfile
kba: c.f. OCR-D/core#368

stweil: It's required here only to provide `add-apt-repository` which is rarely used otherwise.

kba: True, no other project currently uses apt-add-repository. But adding `software-properties-common` pulls in glib libraries, so we could omit the `libglib-2.0.0` dep. Space-wise I doubt that the size difference makes much difference.

Dockerfile
stweil: Fixed now, thank you.

Makefile
```diff
@@ -63,8 +61,6 @@ deps-ubuntu:
 		tesseract-ocr-eng \
 		tesseract-ocr \
 		wget
```
bertsky: We don't need `wget` anymore then, do we?
```suggestion
```

kba: No, plus `wget` comes with ocrd/core.

stweil: Removed now.

Makefile
```diff
@@ -46,25 +44,22 @@ help:
...
 		python3 \
 		python3-pip \
 		libtesseract-dev \
 		libleptonica-dev \
 		tesseract-ocr-eng \
-		tesseract-ocr \
-		wget
-	wget -O - https://github.com/nijel/tesseract-ocr-build/releases/download/4.0.0-1/tesseract.tar.xz | tar -xJf -
-	cp -rt /usr .tesseract/*
+		tesseract-ocr
```
bertsky: by way of https://github.com/OCR-D/ocrd_tesserocr/pull/96#discussion_r353830484, please add 

```suggestion
		tesseract-ocr \
		g++
```

bertsky: Thanks!

stweil: Done. I try to keep such entries ordered, so I did not add it at the end.

Issue #102, opened 2020-01-07
Additional parameter for DPI override


bertsky:
Instead of retransferring https://github.com/OCR-D/core/issues/376: There will have to be a general mechanism in core (sanctioned by spec) at some point in the future (3.0 milestone), but for now (final workshop) we just want a quick solution for Tesseract segmentation and recognition: override the `user_defined_dpi` not with the value from the meta-data density, but a parameter exposed to the user (as a last resort).

Issue #98, opened 2019-12-04
segment-region: crop_polygons creates invalid coordinates


bertsky:
When using `ocrd-tesserocr-segment-region` with `crop_polygons=True`, one will frequently get coordinates extending the segment bbox, which could easily end up in negative coordinates (which is forbidden syntactically in PAGE-XML). 

So maybe Tesseract's [BlockPolygon](https://github.com/tesseract-ocr/tesseract/blob/420cbac876b06beeee271d9f44ba800d943a8a83/src/ccmain/pageiterator.cpp#L378) must be clipped just like its [BoundingBox](https://github.com/tesseract-ocr/tesseract/blob/420cbac876b06beeee271d9f44ba800d943a8a83/src/ccmain/pageiterator.cpp#L331) is clipped?

bertsky:
Also, this parameter should be called just `polygons` (because it is independent of how cropping is done now).

wrznr:
I'd even say the parameter should be called bboxes as soon as this issue is fixed. Polygons should be the default.

bertsky:
> Polygons should be the default.

I agree, but we still have the issue of Tesseract generating invalid (self-intersecting) polygon paths internally, which end up in very strange ways on the consumer side (depending on how the coordinates are being processed, with numpy / skimage / cv2 etc). But maybe it's enough to check against that as well – using `Shapely`, and as a workaround, taking the `exterior` or the self-union...

stweil:
The `tesseract` command line executable also has an [issue with an endless loop](https://github.com/tesseract-ocr/tesseract/issues/2288) when doing segmentation for certain images.

I did not test whether this affects ocrd_tesserocr, too.

bertsky:
Strangely, this affects only image, line and table regions (but not always) – never text regions.

bertsky:
This appears to affect all kinds of regions, but only when they have been rotated internally. Anyway, this is _not_ about clipping to the image/rectangle.

bertsky:
We now have a partial solution in [Tesseract itself](https://github.com/tesseract-ocr/tesseract/pull/2826), but on top of that I still hesitate to make a PR for the convex_hull workaround here...

bertsky:
What if instead of trying to find the bug deep inside Tesseract's polyblk generator we take the liberty of annotating text regions _along with text lines_ in one pass? (Perhaps even with #127 ...)

PR #104, opened 2020-01-10
improve segmentation

### LINKED ISSUE #101: recognize: use PSM_RAW_LINE instead of PSM_SINGLE_LINE
Our OCR-D wrappers have the advantage of allowing to isolate subtasks in a finely grained manner. Tesseract's CLI on the other hand must always provide a good all-in-one compromise, even when called with different PSMs specifically. (E.g. it will always binarize when still necessary for layout analysis, and attempt baseline+xheight+ascender prediction even in `PSM_SINGLE_LINE`.)

Now, for some workflows it might be beneficial to suppress any additional Tesseract-internal segmentation on the provided line images – for instance when that image is cropped and masked from a line polygon already, or clipped already. Under these circumstances, we should rather use `PSM_RAW_LINE`.

But other workflows will just enter with the same rough bounding boxes that Tesseract's CLI would also create internally. Then `PSM_SINGLE_LINE` is a better choice.

So how do we encapsulate this without confusing users, but giving them the best possible results? Do we check the line segment's number of points (for polygon vs bbox workflow), and decide automatically, or expose this as a (thoroghly described) parameter?

@wrznr @kba @stweil

bertsky:
This fixes #101 (using raw_lines by default for textline images, but there are still some corner cases that need to be fixed in Tesseract) and brings a number of segmentation-related improvements:
- interprete `overwrite_regions` more consistently
- annotate `@orientation` (independent of dedicated deskewing processor) for vertical and `@type` for all other text blocks
- no separators and noise regions in reading order
- segment tables into cells and lines so they can be OCRed, too

### DIFF
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -33,7 +33,7 @@
     "ocrd-tesserocr-recognize": {
       "executable": "ocrd-tesserocr-recognize",
       "categories": ["Text recognition and optimization"],
-      "description": "Recognize text lines with Tesseract (using annotated derived images, or masking and cropping images from coordinate polygons)",
+      "description": "Recognize text in lines with Tesseract (using annotated derived images, or masking and cropping images from coordinate polygons)",
       "input_file_grp": [
         "OCR-D-SEG-BLOCK",
         "OCR-D-SEG-LINE",
@@ -70,7 +70,7 @@
      "ocrd-tesserocr-segment-region": {
       "executable": "ocrd-tesserocr-segment-region",
       "categories": ["Layout analysis"],
-      "description": "Segment regions into lines with Tesseract",
+      "description": "Segment page into regions with Tesseract",
       "input_file_grp": [
         "OCR-D-IMG",
         "OCR-D-SEG-PAGE",
@@ -127,7 +127,7 @@
      "ocrd-tesserocr-segment-line": {
       "executable": "ocrd-tesserocr-segment-line",
       "categories": ["Layout analysis"],
-      "description": "Segment page into regions with Tesseract",
+      "description": "Segment regions into lines with Tesseract",
       "input_file_grp": [
         "OCR-D-SEG-BLOCK",
         "OCR-D-GT-SEG-BLOCK"
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -33,7 +33,7 @@
     "ocrd-tesserocr-recognize": {
       "executable": "ocrd-tesserocr-recognize",
       "categories": ["Text recognition and optimization"],
-      "description": "Recognize text in lines with tesseract",
+      "description": "Recognize text lines with Tesseract (using annotated derived images, or masking and cropping images from coordinate polygons)",
       "input_file_grp": [
         "OCR-D-SEG-BLOCK",
         "OCR-D-SEG-LINE",
@@ -48,13 +48,18 @@
         "textequiv_level": {
           "type": "string",
           "enum": ["region", "line", "word", "glyph"],
-          "default": "line",
-          "description": "PAGE XML hierarchy level to add the TextEquiv results to (requires existing layout annotation up to one level above that)"
+          "default": "word",
+          "description": "Lowest PAGE XML hierarchy level to add the TextEquiv results to; when below `region`, implicitly adds segmentation below the line level, but requires existing line segmentation"
         },
         "overwrite_words": {
           "type": "boolean",
           "default": false,
-          "description": "remove existing layout and text annotation below the TextLine level (regardless of textequiv_level)"
+          "description": "Remove existing layout and text annotation below the TextLine level (regardless of textequiv_level)."
+        },
+        "raw_lines": {
+          "type": "boolean",
+          "default": true,
+          "description": "Do not attempt additional segmentation (baseline+xheight+ascenders/descenders prediction) when using line images (i.e. when textequiv_level<region). Disable when line segments/images likely contain components of more than 1 line."
         },
         "model": {
           "type": "string",
@@ -65,7 +70,7 @@
      "ocrd-tesserocr-segment-region": {
       "executable": "ocrd-tesserocr-segment-region",
       "categories": ["Layout analysis"],
-      "description": "Segment regions into lines with tesseract",
+      "description": "Segment regions into lines with Tesseract",
       "input_file_grp": [
         "OCR-D-IMG",
         "OCR-D-SEG-PAGE",
@@ -102,7 +107,7 @@
     "ocrd-tesserocr-segment-line": {
       "executable": "ocrd-tesserocr-segment-line",
       "categories": ["Layout analysis"],
-      "description": "Segment page into regions with tesseract",
+      "description": "Segment page into regions with Tesseract",
       "input_file_grp": [
         "OCR-D-SEG-BLOCK",
         "OCR-D-GT-SEG-BLOCK"
@@ -122,7 +127,7 @@
     "ocrd-tesserocr-segment-word": {
       "executable": "ocrd-tesserocr-segment-word",
       "categories": ["Layout analysis"],
-      "description": "Segment lines into words with tesseract",
+      "description": "Segment lines into words with Tesseract",
       "input_file_grp": [
         "OCR-D-SEG-LINE",
         "OCR-D-GT-SEG-LINE"
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -104,7 +104,27 @@
         }
       }
     },
-    "ocrd-tesserocr-segment-line": {
+     "ocrd-tesserocr-segment-table": {
+      "executable": "ocrd-tesserocr-segment-table",
+      "categories": ["Layout analysis"],
+      "description": "Segment table regions into cell text regions with Tesseract",
+      "input_file_grp": [
+        "OCR-D-SEG-BLOCK",
+        "OCR-D-GT-SEG-BLOCK"
+      ],
+      "output_file_grp": [
+        "OCR-D-SEG-BLOCK"
+      ],
+      "steps": ["layout/segmentation/region"],
+      "parameters": {
+        "overwrite_regions": {
+          "type": "boolean",
+          "default": true,
+          "description": "remove existing layout and text annotation below the region level"
+        }
+      }
+     },
+     "ocrd-tesserocr-segment-line": {
       "executable": "ocrd-tesserocr-segment-line",
       "categories": ["Layout analysis"],
       "description": "Segment page into regions with Tesseract",
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -58,8 +58,8 @@
         },
         "raw_lines": {
           "type": "boolean",
-          "default": true,
-          "description": "Do not attempt additional segmentation (baseline+xheight+ascenders/descenders prediction) when using line images (i.e. when textequiv_level<region). Disable when line segments/images likely contain components of more than 1 line."
+          "default": false,
+          "description": "Do not attempt additional segmentation (baseline+xheight+ascenders/descenders prediction) when using line images (i.e. when textequiv_level<region). Can increase accuracy for certain workflows. Disable when line segments/images may contain components of more than 1 line, or larger gaps/white-spaces."
         },
         "model": {
           "type": "string",
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -194,8 +194,10 @@ class TesserocrRecognize(Processor):
                 line, region_image, region_xywh)
             # todo: Tesseract works better if the line images have a 5px margin everywhere
             tessapi.SetImage(line_image)
-            # RAW_LINE fails with pre-LSTM models, but sometimes better with LSTM models
-            tessapi.SetPageSegMode(PSM.SINGLE_LINE)
+            if self.parameter['raw_lines']:
+                tessapi.SetPageSegMode(PSM.RAW_LINE)
+            else:
+                tessapi.SetPageSegMode(PSM.SINGLE_LINE)
             #if line.get_primaryScript() not in tessapi.GetLoadedLanguages()...
             LOG.debug("Recognizing text in line '%s'", line.id)
             if self.parameter['textequiv_level'] == 'line':
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import os.path
+import itertools
 
 from tesserocr import (
     RIL, PSM,
@@ -141,7 +142,9 @@ class TesserocrRecognize(Processor):
                 #tessapi.SetImage(page_image)
                 
                 LOG.info("Processing page '%s'", page_id)
-                regions = page.get_TextRegion()
+                regions = itertools.chain.from_iterable(
+                    [page.get_TextRegion()] +
+                    [subregion.get_TextRegion() for subregion in page.get_TableRegion()])
                 if not regions:
                     LOG.warning("Page '%s' contains no text regions", page_id)
                 else:
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -369,9 +369,11 @@ def page_update_higher_textequiv_levels(level, pcgts):
     join all first TextEquiv (by the rules governing the respective level)
     into TextEquiv of the next higher level, replacing them.
     '''
-    regions = pcgts.get_Page().get_TextRegion()
+    page = pcgts.get_Page()
     if level != 'region':
-        for region in regions:
+        for region in itertools.chain.from_iterable(
+                [page.get_TextRegion()] +
+                [subregion.get_TextRegion() for subregion in page.get_TableRegion()]):
             lines = region.get_TextLine()
             if level != 'line':
                 for line in lines:
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -384,24 +384,26 @@ class TesserocrRecognize(Processor):
                 result_it.Next(RIL.SYMBOL)
 
 def page_element_unicode0(element):
+    """Get Unicode string of the first text result."""
     if element.get_TextEquiv():
         return element.get_TextEquiv()[0].Unicode
     else:
         return ''
 
 def page_element_conf0(element):
+    """Get confidence (as float value) of the first text result."""
     if element.get_TextEquiv():
         # generateDS does not convert simpleType for attributes (yet?)
         return float(element.get_TextEquiv()[0].conf or "1.0")
     return 1.0
 
 def page_get_reading_order(ro, rogroup):
-    '''Add all elements from the given reading order group to the given dictionary.
+    """Add all elements from the given reading order group to the given dictionary.
     
     Given a dict ``ro`` from layout element IDs to ReadingOrder element objects,
     and an object ``rogroup`` with additional ReadingOrder element objects,
     add all references to the dict, traversing the group recursively.
-    '''
+    """
     if isinstance(rogroup, (OrderedGroupType, OrderedGroupIndexedType)):
         regionrefs = (rogroup.get_RegionRefIndexed() +
                       rogroup.get_OrderedGroupIndexed() +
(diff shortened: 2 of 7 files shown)


stweil:
@bertsky, could you please merge my PR #103 first and then rebase your PR to fix merge conflicts?

ocrd_tesserocr/segment_region.py
```diff
@@ -210,10 +232,19 @@ def _process_page(self, it, page, page_image, page_coords, page_id):
                               # it is a bad idea to create a TextRegion
                               # for it (better set `find_tables` False):
                               # PT.TABLE,
-                              # should actually get a 90° @orientation
-                              # (but that's ultimately for deskewing to decide):
+                              # will also get a 90° @orientation
+                              # (but that can be overriden by deskewing):
```
stweil:
```suggestion
                              # (but that can be overridden by deskewing):
```

bertsky:
> @bertsky, could you please merge my PR #103 first and then rebase your PR to fix merge conflicts?

If it's not urgent, let's please do it the other way round: I have more fixes in the queue here and don't want to resynchronize everything just for cosmetics. Most of your changes are also in my commits here BTW (but that's a total coincidence). We can rebase and merge your PR more easily after this and #102 and #98 are solved.

bertsky:
5b39cc0 fixes the overly simplistic `page_update_higher_textequiv_levels` to propagate text inside tables, too.

c828085 takes that to a whole new level by finally attempting to adhere to `readingDirection`, `textLineOrder`, recursive `ReadingOrder`, and `Relations` of `@type='join'`, e.g. drop-capital vs paragraph or cross-columns paragraph or cross-lines word. 

This should also be covered in the PAGE validator ASAP!

ocrd_tesserocr/recognize.py
```diff
@@ -141,7 +163,9 @@ def process(self):
                 #tessapi.SetImage(page_image)
                 
                 LOG.info("Processing page '%s'", page_id)
-                regions = page.get_TextRegion()
+                regions = itertools.chain.from_iterable(
```
kba: How is this different from just `[page.get_TextRegion()] + [subregion.get_TextRegion() for  subregion in page.get_TableRegion()]`?

kba: Ah, right, it zips the array, so it's `[region1, [subregions of region 1 if any], region2, [subregion of region2 if any]...]`, right?

ocrd_tesserocr/recognize.py
```diff
@@ -194,8 +218,10 @@ def _process_lines(self, tessapi, textlines, region_image, region_xywh):
                 line, region_image, region_xywh)
             # todo: Tesseract works better if the line images have a 5px margin everywhere
             tessapi.SetImage(line_image)
-            # RAW_LINE fails with pre-LSTM models, but sometimes better with LSTM models
-            tessapi.SetPageSegMode(PSM.SINGLE_LINE)
+            if self.parameter['raw_lines']:
```
kba: Will there be a need for another PSM in the future? If so, it might be more intuitive to just follow Tesseracts nomenclature and pass a parameter `PSM` on to Tesseract.

ocrd_tesserocr/recognize.py
```diff
@@ -357,35 +383,147 @@ def _process_glyphs_in_word(self, result_it, word, word_xywh):
...
 
+def page_element_unicode0(element):
+    if element.get_TextEquiv():
+        return element.get_TextEquiv()[0].Unicode
+    else:
+        return ''
+
+def page_element_conf0(element):
+    if element.get_TextEquiv():
+        # generateDS does not convert simpleType for attributes (yet?)
```
kba: This is a regression, they should be converted, I've reopened https://github.com/OCR-D/core/issues/269

ocrd_tesserocr/segment_table.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        overwrite_regions = self.parameter['overwrite_regions']
+        
+        with PyTessBaseAPI(path=TESSDATA_PREFIX) as tessapi:
+            # disable table detection here, so tables will be
+            # analysed as independent text/line blocks:
+            tessapi.SetVariable("textord_tabfind_find_tables", "0")
+            # this should yield additional blocks within the table blocks
+            # from the page iterator, but does not in fact (yet?):
+            #tessapi.SetVariable("textord_tablefind_recognize_tables", "1")
```
kba: Since this is copied here, maybe extend the comment here as well.

ocrd_tesserocr/recognize.py
bertsky: > If so, it might be more intuitive to just follow Tesseracts nomenclature and pass a parameter `PSM` on to Tesseract.

...instead of `textequiv_level` you mean? The problem is that Tesseract's PSMs are already non-orthogonal and more fine-grained than our/PAGE's hierarchy. Plus we should strive to encapsulate Tesseract's idiosyncrasies here. 

This new distinction is necessary though, because for LSTM models, PSM.SINGLE_LINE can be detrimental, esp. on images that have already been masked from polygonal line segmentation. Moreover, with current Tesseract, this can still crash under rare circumstances. So this depends on models and data, and we should at least provide the old PSM as an opt-out.

ocrd_tesserocr/recognize.py
bertsky: But it's like this ever since 1.0.1 – the last working previous version (float conversion) was 1.0.b8.

Moreover, IIRC generateDS documentation says explicitly this is a _won't fix_...

Are you certain this breaks round-trip?

ocrd_tesserocr/recognize.py
kba: > ...instead of `textequiv_level` you mean?

Instead of `raw_lines`. But you're right, this is about `RAW_LINE and `SINGLE_LINE` and the parameter is perfectly expressive for that.

kba:
> This should also be covered in the PAGE validator ASAP!

@tboenig What GT do we have which uses these features and is robust enough for testing?

ocrd_tesserocr/recognize.py
bertsky: No, the opposite, it's meant to flatten the lists.

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -33,7 +33,7 @@
     "ocrd-tesserocr-recognize": {
       "executable": "ocrd-tesserocr-recognize",
       "categories": ["Text recognition and optimization"],
-      "description": "Recognize text in lines with tesseract",
+      "description": "Recognize text lines with Tesseract (using annotated derived images, or masking and cropping images from coordinate polygons)",
```
wrznr: For me, *text in lines* is less ambig than *text lines* which may be misunderstood as detecting lines.

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -48,13 +48,18 @@
         "textequiv_level": {
           "type": "string",
           "enum": ["region", "line", "word", "glyph"],
-          "default": "line",
-          "description": "PAGE XML hierarchy level to add the TextEquiv results to (requires existing layout annotation up to one level above that)"
+          "default": "word",
```
wrznr: @stweil The change of the default level is here!

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -65,7 +70,7 @@
      "ocrd-tesserocr-segment-region": {
       "executable": "ocrd-tesserocr-segment-region",
       "categories": ["Layout analysis"],
-      "description": "Segment regions into lines with tesseract",
+      "description": "Segment regions into lines with Tesseract",
```
wrznr: This not correct, right? The descriptions of `segment-region` and `segment-line` are interchanged. (We have to get rid of the processor suffixes.)

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -99,10 +104,30 @@
...
+          "default": true,
+          "description": "remove existing layout and text annotation below the region level"
+        }
+      }
+     },
+     "ocrd-tesserocr-segment-line": {
       "executable": "ocrd-tesserocr-segment-line",
       "categories": ["Layout analysis"],
-      "description": "Segment page into regions with tesseract",
+      "description": "Segment page into regions with Tesseract",
```
wrznr: Cf. above.

ocrd_tesserocr/recognize.py
```diff
@@ -357,35 +383,147 @@ def _process_glyphs_in_word(self, result_it, word, word_xywh):
...
+        return float(element.get_TextEquiv()[0].conf or "1.0")
+    return 1.0
+
+def page_get_reading_order(ro, rogroup):
+    '''Add all elements from the given reading order group to the given dictionary.
```
wrznr: Double quotes are used elsewhere.

ocrd_tesserocr/recognize.py
```diff
@@ -357,35 +383,147 @@ def _process_glyphs_in_word(self, result_it, word, word_xywh):
...
+        regionrefs = (rogroup.get_RegionRef() +
+                      rogroup.get_OrderedGroup() +
+                      rogroup.get_UnorderedGroup())
+    for elem in regionrefs:
+        ro[elem.get_regionRef()] = elem
+        if not isinstance(elem, (RegionRefType, RegionRefIndexedType)):
+            page_get_reading_order(ro, elem)
+        
 def page_update_higher_textequiv_levels(level, pcgts):
     '''Update the TextEquivs of all PAGE-XML hierarchy levels above ``level`` for consistency.
```
wrznr: See above.

ocrd_tesserocr/recognize.py
kba: > Are you certain this breaks round-trip?

Not necessarily but what's the point of having types in XSD if you don't validate data and don't produce native types when code generating. It's not high in the priority list though. But when I re-generate the next time (there's also the issue with building with parent element refs etc.) I'll have a look.

ocrd_tesserocr/ocrd-tool.json
kba: True, you might read that as "line segmentation"

ocrd_tesserocr/recognize.py
bertsky: I am with you on this, I was puzzled about this decision myself. But maybe it's just a matter of resources and will change in the future. (Also, remember this only affects atomic attribute types.)

ocrd_tesserocr/ocrd-tool.json
bertsky: True! We wanted to fix that a long time ago!

bertsky:
> In addition, I know it is sometimes inconvenient but would smaller PRs be an option for you? In this case, IMHO RO handling and table segmentation could have been split up into separate feature-related PRs.

Sure, I can split the last commit as a separate PR, but that would of course have to be based on this PR anyway. (I still have not received an answer from anyone whether this is better.) Just for my understanding: you don't want to see and comment the diffs accumulated over all commits in `files changed`? (If that is so, why not view/comment the `commits` individually?)

wrznr:
Because these commits are not independent. I review the first commit, encounter a problem and go nuts for nothing because you have fixed that problem in a later commit. Feature-based PRs make it much easier to understand the internals of what you are proposing. As it stands, I can only do a formal review wrt. syntax and language issues.

>  I can split the last commit as a separate PR

Pls. do not but maybe keep your PRs smaller in the future.

> (I still have not received an answer from anyone whether this is better.)

It is. Okay?

ocrd_tesserocr/ocrd-tool.json
kba: Yeah, that was wrong from the first commit... 😊

ocrd_tesserocr/recognize.py
kba: For the docstring? IIUC single and double quotes are exactly the same for triple-quote-delimited strings.

kba:
> but that would of course have to be based on this PR anyway
> ...
> Because these commits are not independent.

We all appreciate concise PR but in my experience with splitting up PR into smaller dependent chunks is often very frustrating because you have to maintain multiple branches and while the "dependency" PR are not merged, the diff of the PR is against master branch so it's even more confusing and redundant.

kba:
>> @bertsky, could you please merge my PR #103 first and then rebase your PR to fix merge conflicts?

> If it's not urgent, let's please do it the other way round

I would not call it urgent but I also don't see why this should block #103 . Just a few conflicts in ocrd-tool.json, feel free to pull https://github.com/kba/ocrd_tesserocr/pull/new/raw-line if you don't want to solve them yourself.

And I try to avoid rebasing as much as possible. Merge commits might "pollute" the log but there's always `git log --no-merges` and a consistent state across forks.

bertsky:
@wrznr @kba thanks for clarifying! Let's please discuss this further just once:

> We all appreciate concise PR but in my experience with splitting up PR into smaller dependent chunks is often very frustrating because you have to maintain multiple branches and while the "dependency" PR are not merged, the diff of the PR is against master branch so it's even more confusing and redundant.

Okay, I'm glad we all agree on that.

> Because these commits are not independent. I review the first commit, encounter a problem and go nuts for nothing because you have fixed that problem in a later commit.

I always try to make sequences of compact commits, using rebasing, which can be understood. The tendency is rather towards too large commits for me. So I am surprised to read it's the opposite in your view. Which change(s) did you find difficult to follow here?

Also, if it is this way around, why not use the accumulated diff perspective instead?

> Feature-based PRs make it much easier to understand the internals of what you are proposing.

I agree this PR can be said to combine impovements across several _features_. (That's why I have used several commits.) But they all relate to segmentation, and I digged them up together by testing and Tesseract debugging. Are you serious I should separate them into individual branches, switching between branches while testing workspaces? How would you then test these yourself?

> but maybe keep your PRs smaller in the future.

It boils down to this: non-trivial changes necessitate larger PRs IMO. But I will try!

bertsky:
> I would not call it urgent but I also don't see why this should block #103

As I said: I have got more (stashed/uncommited) changes in the queue, and this would mean more synchronization work for me. I don't see why I should pay that (unnecessary) price.

stweil:
I tested this PR with one of our old book (https://digi.bib.uni-mannheim.de/urn/urn:nbn:de:bsz:180-digad-22977) using this process:

    time -p ocrd process -m mets.xml \
      "tesserocr-segment-region -I MAX -O OCR-D-SEG-BLOCK -p '{\"crop_polygons\": true}'" \
      "tesserocr-segment-line -I OCR-D-SEG-BLOCK -O OCR-D-SEG-LINE" \
      "tesserocr-recognize -I OCR-D-SEG-LINE -O OCR-D-OCR-TESS -p '{\"model\": \"Fraktur_5000000_0.502_198857\"}'"

The results with the previous code and without `crop_polygons` were better in most cases where there was a difference. This might be caused by the use of raw lines.

bertsky:
@stweil thanks for testing!

However, I think there is a misunderstanding here: this PR is not about `crop_polygons` – not at all. As I have [already stated](https://github.com/OCR-D/ocrd_tesserocr/issues/98#issuecomment-567391635), my [fixes](https://github.com/tesseract-ocr/tesseract/pull/2826) for `BlockPolygon` are not enough to make this usable yet. (And when they do become ready, we agreed to rename the option and make it default.)

So, please repeat your test with `crop_polygons=false` (the default)!

Also, if you do still find regressions, could you please use ocrd-dinglehopper to compare before/after and post the relevant pages here?

(To embed the HTML, you can use the https://jsbin.com service – just paste the HTML on the left, then click `Share`, `Output-only`, and `Link`.)

wrznr:
@bertsky First and foremost, I am gracious for your work and it is up to you how you structure your commits and PRs. When I ask for smaller PRs then it is a **personal** preference (and best practice in many, many companies and projects) which helps me to provide substantial reviews. The sheer amount of changes and code additions in this PR is not trackable for me, so my review is, again, purely formal. (@kba I also doubt that merging is any faster and less frustrating than for multiple smaller PRs.)

ocrd_tesserocr/recognize.py
bertsky: So, do you want me to change it, or resolved?

bertsky:
@kba updating version string in tool json and setup: Will you do that after the merge, along with release tag? How about 0.7?

ocrd_tesserocr/recognize.py
wrznr: Despite the semantics, I think we should consistent at least within single files. So pls. change.

ocrd_tesserocr/ocrd-tool.json
bertsky: cf 
8064c0a

ocrd_tesserocr/ocrd-tool.json
bertsky: cf 
8064c0a

ocrd_tesserocr/segment_table.py
bertsky: cf 
66da5f6

ocrd_tesserocr/segment_region.py
bertsky: cf 
66da5f6

ocrd_tesserocr/recognize.py
bertsky: cf 
66da5f6

ocrd_tesserocr/recognize.py
bertsky: cf 
66da5f6

bertsky:
Sorry @wrznr bad timing. Just added 68a49485739 to complete the PR (removing a `FIXME`). 

What do you think: Is it okay to import `page_get_reading_order` from another processor as long as we don't have these in ocrd_utils?

wrznr:
Since the processor is from the same package, I think it is okay to do so. (Maybe with the corresponding issue in core and a Fixme comment?)

kba:
> Since the processor is from the same package

Even if it were not, as long as that's properly documented and installable. Better to avoid code duplication and have a slightly more involved setup.

PR to ocrd_utils of course always welcome.

kba:
> @kba updating version string in tool json and setup: Will you do that after the merge, along with release tag (or should I add it to the PR)? How about 0.7?

Sure, let me know when it's ready.

bertsky:
> PR to ocrd_utils of course always welcome.

I will make an attempt. Have to think of the general API though (RO iterator, modification, region type filters) and harmonize with ideas from ocrd_segment.

bertsky:
> > @kba updating version string in tool json and setup: Will you do that after the merge, along with release tag (or should I add it to the PR)? How about 0.7?
> 
> Sure, let me know when it's ready.

It's ready alright. Of course – again – some extra effort should be made to add test coverage, but I don't have the time now.

@stweil have you run your test to satisfaction?

bertsky:
@stweil ok, thanks – could you please upload the original file of that page here, so I can have a look myself?

stweil:
> could you please upload the original file of that page here

@bertsky, is the [METS file](https://digi.bib.uni-mannheim.de/mets/urn:nbn:de:bsz:180-digad-22977) with all file links sufficient, or do you need other files, too? I tested physical page 7.

bertsky:
@stweil this was sufficient, thanks.

> and still see a regression with the new code. For the page shown above, the new code does not recognize the first line ("Vorred ...") and misses an "a" in line 5 ("das"). The rest of that page is identical now. Other pages show similar small differences, and in most cases the old code gave getter results.

I see similar problems with other Fraktur models.

Yet, this is to be expected: your workflow is too simplistic for `raw_lines=true`:
- ocrd-tesserocr-segment-line gives only bboxes, not polygons (as would ocrd-cis-ocropy-segment, or ocrd-tesserocr-segment-line+ocrd-cis-ocropy-resegment)
- ocrd-tesserocr-segment-line does not suppress foreground components of neighbouring/intruding regions (as would ocrd-cis-ocropy-clip)

Now, I already documented this in the JSON description of `raw_lines`:
> Do not attempt additional segmentation (baseline+xheight+ascenders/descenders prediction) when using line images (i.e. when textequiv_level<region). Disable when line segments/images likely contain components of more than 1 line.

So to me this is a _wontfix_ issue. Unless you would like to have more documentation, perhaps in README.md or the description of ocrd-tesserocr-segment-line?

ocrd_tesserocr/segment_region.py
```diff
@@ -71,7 +74,9 @@ def process(self):
                 tessapi.SetVariable("textord_tabfind_find_tables", "1") # (default)
                 # this should yield additional blocks within the table blocks
                 # from the page iterator, but does not in fact (yet?):
-                tessapi.SetVariable("textord_tablefind_recognize_tables", "1")
+                # (and it can run into assertion errors when the table structure
+                #  does not meet certain homogenity expectations)
```
stweil:
```suggestion
                #  does not meet certain homogeneity expectations)
```

bertsky:
> Yet, this is to be expected: your workflow is too simplistic for `raw_lines=true`:
> So to me this is a _wontfix_ issue. Unless you would like to have more documentation, perhaps in README.md or the description of ocrd-tesserocr-segment-line?

@stweil Another [problem](https://github.com/OCR-D/ocrd_tesserocr/issues/101#issuecomment-576926456 ) for raw lines just came up, which is next to independent of workflow configuration. This has swung the decision in favour of the old behaviour as default. 

I can still remove the new mode entirely, if this is what everyone wants.

bertsky:
@kba I think you can merge and release now.

kba:
Thanks, released as [v0.7.0](https://pypi.org/project/ocrd-tesserocr/0.7.0/)

Issue #107, opened 2020-01-21
recognize: expose character white/blacklisting parameters


bertsky:
We should pass the following additional parameters/variables from ocrd-tool.json to Tesseract's API:

param | description
----- | ------
`tessedit_char_whitelist` | Whitelist of chars to recognize
`tessedit_char_blacklist` | Blacklist of chars not to recognize
`tessedit_char_unblacklist` | List of chars to override tessedit_char_blacklist

(But maybe omit the `tessedit_` prefix.)

PR #108, opened 2020-01-23
all: add dpi parameter as manual override to image metadata

### LINKED ISSUE #102: Additional parameter for DPI override

bertsky:
Fixes #102. 

This does influence segmentation and OSD a lot, but not so sure about (LSTM) recognition. Maybe we should document this somewhere – it's one of very few parameters we have to influence the quality of Tesseract's layout analysis (i.e. deliberately setting a fake value can improve results).

Note that Tesseract is optimised for modern fonts (which are typically smaller than historic ones) and thus biased against historic prints. So setting higher than factual DPI could be a useful recommendation for block (and maybe even line) segmentation. But this is still conjecture and I have only little evidence supporting it – someone would have to make systematic measurements first.

### DIFF
--- a/ocrd_tesserocr/crop.py
+++ b/ocrd_tesserocr/crop.py
@@ -94,10 +94,18 @@ class TesserocrCrop(Processor):
                     # image must not have been rotated or cropped already,
                     # abort if no such image can be produced:
                     feature_filter='deskewed,cropped')
-                if page_image_info.resolution != 1:
+                if self.parameter['dpi'] > 0:
+                    dpi = self.parameter['dpi']
+                    LOG.info("Page '%s' images will use %d DPI from parameter override", page_id, dpi)
+                elif page_image_info.resolution != 1:
                     dpi = page_image_info.resolution
                     if page_image_info.resolutionUnit == 'cm':
                         dpi = round(dpi * 2.54)
+                    LOG.info("Page '%s' images will use %d DPI from image meta-data", page_id, dpi)
+                else:
+                    dpi = 0
+                    LOG.info("Page '%s' images will use DPI estimated from segmentation", page_id)
+                if dpi:
                     tessapi.SetVariable('user_defined_dpi', str(dpi))
                     zoom = 300 / dpi
                 else:
--- a/ocrd_tesserocr/deskew.py
+++ b/ocrd_tesserocr/deskew.py
@@ -100,11 +100,20 @@ class TesserocrDeskew(Processor):
                     # (we will overwrite @orientation anyway,)
                     # abort if no such image can be produced:
                     feature_filter='deskewed' if oplevel == 'page' else '')
-                if page_image_info.resolution != 1:
+                if self.parameter['dpi'] > 0:
+                    dpi = self.parameter['dpi']
+                    LOG.info("Page '%s' images will use %d DPI from parameter override", page_id, dpi)
+                elif page_image_info.resolution != 1:
                     dpi = page_image_info.resolution
                     if page_image_info.resolutionUnit == 'cm':
                         dpi = round(dpi * 2.54)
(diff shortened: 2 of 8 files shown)


ocrd_tesserocr/crop.py
```diff
@@ -94,10 +94,18 @@ def process(self):
                     # image must not have been rotated or cropped already,
                     # abort if no such image can be produced:
                     feature_filter='deskewed,cropped')
-                if page_image_info.resolution != 1:
+                if self.parameter['dpi'] > 0:
+                    dpi = self.parameter['dpi']
+                    LOG.info("Page '%s' images will use %d DPI from override", page_id, dpi)
```
stweil:
```suggestion
                    LOG.info("Page '%s' images will use %d DPI from parameter", page_id, dpi)
```

kba: Please commit, I lack the rights.

ocrd_tesserocr/deskew.py
```diff
@@ -100,11 +100,20 @@ def process(self):
                     # (we will overwrite @orientation anyway,)
                     # abort if no such image can be produced:
                     feature_filter='deskewed' if oplevel == 'page' else '')
-                if page_image_info.resolution != 1:
+                if self.parameter['dpi'] > 0:
+                    dpi = self.parameter['dpi']
+                    LOG.info("Page '%s' images will use %d DPI from override", page_id, dpi)
```
kba: same as above. "parameter" or "parameter override" is preferable.

ocrd_tesserocr/recognize.py
```diff
@@ -155,12 +155,19 @@ def process(self):
                                                 for name in self.parameter.keys()])]))
                 page_image, page_xywh, page_image_info = self.workspace.image_from_page(
                     page, page_id)
-                if page_image_info.resolution != 1:
+                if self.parameter['dpi'] > 0:
+                    dpi = self.parameter['dpi']
+                    LOG.info("Page '%s' images will use %d DPI from override", page_id, dpi)
```
kba: as above

ocrd_tesserocr/deskew.py
bertsky: Chose the latter and force-pushed.

bertsky:
@kba when you merge, please don't forget to update `CHANGELOG.md` this time.

As for versioning, I think 108, 109 and 110 are all new features, therefore I suggest `0.8.0`.

kba:
> please don't forget to update CHANGELOG.md this time

I'll try to, but I'm grateful for contributions. You have the best overview to outline your PR in broad strokes. Updated for 0.7.0 in master.

bertsky:
> I'll try to, but I'm grateful for contributions. You have the best overview to outline your PR in broad strokes.

ok, first, for 0.7.0, I believe these need to move from `Added` to `Changed`:
- interprete overwrite_regions more consistently, #104
- annotate @orientation (independent of dedicated deskewing processor) for vertical and @type for all other text blocks, #104
- no separators and noise regions in reading order, #104

then for 0.8.0, how about this:
- Added:
  - parameter `dpi` overriding pixel density meta-data in all processors, #102 
  - parameters `char_whitelist`, `char_blacklist`, `char_unblacklist` in `recognize`, #107 
  - set `lstm_choice_mode=2` when `textequiv_level=glyph` to get character alternatives from LSTM in `recognize`, #7

PR #109, opened 2020-01-24
recognize: add char white/un/blacklisting parameters

### LINKED ISSUE #107: recognize: expose character white/blacklisting parameters

bertsky:
Fixes #107.

Can be quite useful when looking at the unicharset, e.g.:
```sh
combine_tessdata -e /usr/local/share/tessdata/Fraktur.traineddata Fraktur.lstm-unicharset
less Fraktur.lstm-unicharset
ocrd-tesserocr-recognize -I OCR-D-SEG-LINE -O OCR-D-OCR-TESS -p '{"char_blacklist": "_[]{}|®©@€"}'
```

I am not so sure how to handle **normalization** for historic texts, though. We could offer Unicode `NFC` and `NFKC`, but they don't fit so well into our GT transcription level scheme (for which we still have no implementation BTW; if have only [started](https://github.com/ASVLeipzig/cor-asv-ann/blob/39cae08964cce754471248065a89a48808fd43e9/ocrd_cor_asv_ann/lib/alignment.py#L302) this for my module). But the standard models do contain characters that we don't want in levels 1 and 2, e.g.:
- `ﬀ`
- `ﬆ`
- `ﬄ`
- `ﬅ`

### DIFF
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -61,6 +61,21 @@
           "default": false,
           "description": "Do not attempt additional segmentation (baseline+xheight+ascenders/descenders prediction) when using line images (i.e. when textequiv_level<region). Can increase accuracy for certain workflows. Disable when line segments/images may contain components of more than 1 line, or larger gaps/white-spaces."
         },
+        "char_whitelist": {
+          "type": "string",
+          "default": "",
+          "description": "Enumeration of character hypotheses (from the model) to allow exclusively; overruled by blacklist if set."
+        },
+        "char_blacklist": {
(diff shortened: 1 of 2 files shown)


stweil:
A testcase would be nice (and satisfy codecov).

PR #126, opened 2020-05-12
Recognize with padding

bertsky:
This was on our to-do list for a long time – I don't know just why we did not just do it.

This is for cases where the line polygon is quite close to the foreground and there is no `AlternativeImage` with some padding around the margins (as `ocrd-cis-ocropy-dewarp` would yield). AFAICT Tesseract needs some margin (I'd say at least 5px) for good recognition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,15 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [0.8.3] - 2020-05-12
+
 Fixed:
 
   * recognize: ignore empty RO group
+
+Changed:
+
+  * recognize: add `padding` parameter
 
 ## [0.8.2] - 2020-04-08
 
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -68,6 +68,12 @@
           "default": false,
           "description": "Remove existing layout and text annotation below the TextLine level (regardless of textequiv_level)."
         },
+        "padding": {
+          "type": "number",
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,10 +5,16 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [0.8.3] - 2020-05-12
```
kba, line 8: I follow the convention to only change version numbers when releasing and adding changes under the `Unreleased` headline.

kba:
> Looks good! Shall I release 0.8.3?

No need, you are owner of the project on PyPI, so release as you see fit :)

CHANGELOG.md
bertsky: Right, but this is a hen and egg problem, isn't it? Or do you make the change to the version number/changelog directly on master?

bertsky:
> No need, you are owner of the project on PyPI, so release as you see fit :)

damn, now I got the order confused. (PyPI release before merge/GH release.)

CHANGELOG.md
kba: Yeah, I make the version changes directly in master:

* Update changelog
* Update setup.py (or package.json, Makefile `VERSION` macro, `.version` or whatever the scheme)
* Commit as `:package: v0.0.0`
* git tag v0.0.0
* git push; git push --tags
* Release to PyPI, npm, nexus etc.
* Create a release on GitHub based on the tag.

For more complex setups (like core's mult-module setup), I have a little cheatsheet so I don't forget the steps.

PR #128, opened 2020-05-20
segment-region: clip polygons to parent...

bertsky:
(before annotating polygon coordinates, ensure they fit
 into their parent if it exists, or the page frame otherwise)

This happens for example when the page was deskewed and the region is already on the edge of the (rotated, enlarged) page frame. So the result would be syntactically invalid. 

This is another (previously undetected) instance of https://github.com/OCR-D/core/issues/489, with the [same solution](https://github.com/cisocrgroup/ocrd_cis/pull/48/commits/01638c45d6b777adccf03ad457e2b75909ea107f).

### DIFF
--- a/ocrd_tesserocr/segment_region.py
+++ b/ocrd_tesserocr/segment_region.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import os.path
+from shapely.geometry import Polygon
 from tesserocr import (
     PyTessBaseAPI,
     PSM, RIL, PT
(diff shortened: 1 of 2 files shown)


ocrd_tesserocr/segment_region.py
```diff
@@ -299,3 +302,35 @@ def _process_page(self, it, page, page_image, page_coords, page_id):
             not og.get_UnorderedGroupIndexed()):
             # schema forbids empty OrderedGroup
             ro.set_OrderedGroup(None)
+
+def polygon_for_parent(polygon, parent):
+    """Clip polygon to parent polygon range.
+    
+    (Should be moved to ocrd_utils.coordinates_for_segment.)
```
kba, line 309: :+1:

bertsky:
@kba could you please merge this as well and make another release? (We can take it out again as soon as this is handled by core, but given the many recent and coming changes to the PAGE API, the rewrite will probably be rather large and take more time.)

kba:
Sure

PR #132, opened 2020-06-24
crop: convert from relative to absolute

bertsky:
- respect previous reflect/rotate by converting
  back Border polygon with inverse transform
- allow running on deskewed PAGE/image
- clip Border polygon to original image frame
 (before annotating polygon coordinates, ensure they fit
  into the page frame)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Fixed:
+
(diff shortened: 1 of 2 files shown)


bertsky:
Note: the cropper here still has its merits over ocrd-anybaseocr-crop. It cannot cope with textual noise (facing pages), but otherwise it's more robust than the latter.

bertsky:
>     * allow running on deskewed PAGE/image

This also means there is a dependency on https://github.com/OCR-D/core/pull/525

PR #138, opened 2020-08-13
pip install -U pip before pip install

kba:
Let's see whether that fixes the issue. I guess opencv-headless required skbuild but dropped that requirement recently, but haven't checked yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,14 +64,17 @@ deps-ubuntu:
 
 # Install python deps via pip
 deps:
+	$(PIP) install -U pip
 	$(PIP) install -r requirements.txt
 
 # Install testing python deps via pip
 deps-test:
+	$(PIP) install -U pip
 	$(PIP) install -r requirements_test.txt
 
 # Install
 install:
+	$(PIP) install -U pip
 	$(PIP) install .
 
 # Build docker image
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,3 +1,2 @@
 pytest >= 4.4.0
 coverage >= 4.5.2
-scikit-build
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,2 +1,3 @@
 pytest >= 4.4.0
 coverage >= 4.5.2
+scikit-build


kba:
Definitely an issue with opencv. Adding `scikit-build` fixed that particular issue but now throws this:

```
Installing collected packages: pyrsistent, jsonschema, click, lxml, numpy, Pillow, ocrd-utils, ocrd-models, ocrd-modelfactory, bagit, shapely, urllib3, certifi, chardet, requests, bagit-profile, pyyaml, ocrd-validators, atomicwrites, MarkupSafe, Jinja2, itsdangerous, Werkzeug, Flask, wrapt, Deprecated, opencv-python-headless, ocrd, tesserocr
  Running setup.py install for opencv-python-headless ... - error
    Complete output from command /usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-fj9bt_9w/opencv-python-headless/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-uwpfxbcp-record/install-record.txt --single-version-externally-managed --compile:
    /usr/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'
      warnings.warn(msg)
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/skbuild/setuptools_wrap.py", line 560, in setup
        cmkr = cmaker.CMaker(cmake_executable)
      File "/usr/local/lib/python3.6/dist-packages/skbuild/cmaker.py", line 95, in __init__
        self.cmake_version = get_cmake_version(self.cmake_executable)
      File "/usr/local/lib/python3.6/dist-packages/skbuild/cmaker.py", line 82, in get_cmake_version
        "Problem with the CMake installation, aborting build. CMake executable is %s" % cmake_executable)
    
    Problem with the CMake installation, aborting build. CMake executable is cmake
    
    ----------------------------------------
```

Probably should bisect recent opencv-headless versions to find the culprit.

kba:
:fire: eek, this is now happening for core as well (but only py35 and py36). https://app.circleci.com/pipelines/github/OCR-D/core/625/workflows/cae742df-a08b-4a16-a569-58becbb38c55/jobs/1607

bertsky:
> Let's see whether that fixes the issue. I guess opencv-headless required skbuild but dropped that requirement recently, but haven't checked yet.

Can confirm that. `opencv-python`'s setup.py used to install its `scikit-build` on the fly (by calling `pip` cmdline). Now it outright imports it on the module level, cf [here](https://github.com/skvark/opencv-python/blob/d356af12a31fbcb6b7a82372555a7569639bb483/setup.py#L10). 

Don't know how that is supposed to work, but: apparently we are _not_ required to install this ourselves. Instead, [this issue](https://github.com/skvark/opencv-python/issues/372) says our `pip` is simply too old.

> /usr/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'

To me that's rather indicative of an outdated setuptools or pip.

Which reminds me that nowhere in our CI setup we update them (we only fetch the most recent system packages).

So I suggest rolling back the `scikit-build` requirement and instead introduce an `pip install -U pip` step in `deps`.

bertsky:
> ek, this is now happening for core as well (but only py35 and py36).

Same diagnosis IMO: In core we do have a `pip install -U pip` step, but as part of `install`'s recipe, not for `deps`. It must be the first thing we do in the whole chain.

kba:
> So I suggest rolling back the `scikit-build` requirement and instead introduce an `pip install -U pip` step in `deps`.

That fixed it :) Thanks! https://app.circleci.com/pipelines/github/OCR-D/ocrd_tesserocr/255/workflows/6604b13c-774e-4cf7-98df-26c8f937de41/jobs/248

skvark:
> Can confirm that. `opencv-python`'s setup.py used to install its `scikit-build` on the fly (by calling `pip` cmdline). Now it outright imports it on the module level, cf [here](https://github.com/skvark/opencv-python/blob/d356af12a31fbcb6b7a82372555a7569639bb483/setup.py#L10).
> 
> Don't know how that is supposed to work, but: apparently we are _not_ required to install this ourselves. Instead, [this issue](https://github.com/skvark/opencv-python/issues/372) says our `pip` is simply too old.

Hello, noticed this issue via linked issues. For clarity, ``scikit-build`` gets installed via ``pyproject.toml`` which is the new way of defining for example build systems for Python packages. See: https://snarky.ca/what-the-heck-is-pyproject-toml/

You were probably using the old ``manylinux1`` pre-built wheels earlier. There wasn't source distribution in PyPI then so everything worked fine. After ``manylinux2014`` was introduced along with the source distributions, old version of ``pip`` bailed out on the pre-built wheels because it doesn't understand that new standard. Then it tried to compile from source with the new source distribution. However, that failed also due to too old ``pip`` version because it didn't understand the build system definition in ``pyproject.toml``.

kba:
Thanks @skvark for the explanation, much appreciated. The skbuild problem forced us to update pip which is a good thing. Also good to know about `pyproject.toml`, we will discuss whether it might be worth switching.

PR #140, opened 2020-08-15
Remove blanks at line endings

stweil:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ Fixed:
   * segment-region: add `sparse_text` mode choice
   * segment-line: make intersection with parent more robust
   * segment-table: use `SPARSE_TEXT` mode for cells
-  
+
 Changed:
 
   * Depend on OCR-D/core v2.4.4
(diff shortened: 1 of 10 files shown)


stweil:
We cannot avoid handling this problem: good editors remove such blanks automatically, so any time I edit one of those files, I am confronted with the problem, and so I am when I run `git diff` which shows the blanks considered wrong in red. That's why I prefer to clean the code in one step (and maybe add a git rule which prevents future blanks at line endings).

stweil:
I still think it would be good to follow the convention and avoid blanks at line endings, but close this issue to reduce the list of open pull requests.

bertsky:
> I still think it would be good to follow the convention and avoid blanks at line endings, but close this issue to reduce the list of open pull requests.

Yes, in the meantime I have seen [these arguments](https://github.com/PyCQA/pylint/issues/1368), i.e. pylint does not support `no-space-check=empty-line` anymore anyway. So I agree now we should at least phase out the old convention. (I'll try to adapt my editor accordingly.)

Issue #139, opened 2020-08-15
Processor terminates with exception: ocrd-tesserocr-segment-region exited with non-zero


stweil:
Physical page 129 of http://nbn-resolving.de/urn:nbn:de:bsz:180-digad-10044 raises a fatal exception in the standard [workflow for slower processors](https://ocr-d.de/en/workflows):

    20:13:52.948 INFO processor.TesserocrSegmentRegion - INPUT FILE 128 / IMG_OCR-D-BIN-DENOISE-DESKEW_460017
    20:13:52.971 INFO processor.TesserocrSegmentRegion - Page 'IMG_OCR-D-BIN-DENOISE-DESKEW_460017' images will use 300 DPI from image meta-data
    20:13:52.971 INFO processor.TesserocrSegmentRegion - Detecting regions in page 'IMG_OCR-D-BIN-DENOISE-DESKEW_460017'
    20:13:53.352 INFO processor.TesserocrSegmentRegion - Detected region 'region0000': 1425,976 1421,976 1424,2217 1425,2217 (HORZ_LINE)
    20:13:53.353 INFO processor.TesserocrSegmentRegion - Detected region 'region0001': 1425,1788 1423,976 1413,976 1416,2194 1425,2194 (HORZ_LINE)
    20:13:53.354 INFO processor.TesserocrSegmentRegion - Detected region 'region0002': 523,97 525,967 958,966 956,97 (TABLE)
    20:13:53.354 INFO processor.TesserocrSegmentRegion - Detected region 'region0003': 392,292 432,292 434,1160 394,1160 (FLOWING_TEXT)
    20:13:53.355 INFO processor.TesserocrSegmentRegion - Detected region 'region0004': 378,162 401,162 401,307 378,307 (FLOWING_TEXT)
    20:13:53.355 INFO processor.TesserocrSegmentRegion - Detected region 'region0005': 92,110 330,109 332,1157 94,1158 (TABLE)
    Traceback (most recent call last):
      File "/home/stweil/src/github/OCR-D/venv-20200815/bin/ocrd-tesserocr-segment-region", line 8, in <module>
        sys.exit(ocrd_tesserocr_segment_region())
      File "/home/stweil/src/github/OCR-D/venv-20200815/lib/python3.7/site-packages/click/core.py", line 829, in __call__
        return self.main(*args, **kwargs)
[...]

PR #141, opened 2020-08-15
Don't raise exception for empty intersection (fix issue #139)

stweil:

### DIFF
--- a/ocrd_tesserocr/segment_region.py
+++ b/ocrd_tesserocr/segment_region.py
@@ -325,7 +325,8 @@ def polygon_for_parent(polygon, parent):
     interp = childp.intersection(parentp)
     if interp.is_empty:
         # FIXME: we need a better strategy against this
-        raise Exception("intersection of would-be segment with parent is empty")
+        LOG.error("intersection of would-be segment with parent is empty")
+        return polygon
     if interp.type == 'MultiPolygon':
         interp = interp.convex_hull
     return interp.exterior.coords[:-1] # keep open


kba:
But if the child element is outside the parent element, that is an error that should be adressed. I agree that raising an exception shifts the responsibility for debugging to the user, that is not ideal. But this should be debugged nonetheless.

stweil:
> But this should be debugged nonetheless.

Sure, that's why I added `LOG.error`. But you are right, this is not a fix but a workaround, so issue #139 should be kept open even with this commit.

stweil:
> But if the child element is outside the parent element, that is an error that should be adressed.

Maybe both parent and child are empty, then the child is not outside, but the same error is triggered.

bertsky:
Introducing this exception was a conscious decision for the current state of coordinate consistency in OCR-D.

We know that most processors currently write illegal coordinates sometimes, which ultimately fails at validation time with lots of hard to debug or trace or visualise error messages. But shifting the burden for correct coordinates from validation to implementation is a large effort, and we need core to support this as much as possible. That's where [this idea](https://github.com/OCR-D/core/issues/489#issuecomment-634731069) will step in.

But looking at the example in #139 I can now see that this will happen much more frequently than I thought in ocrd_tesserocr. So yes, we need to offer some workaround as well. I'll have to think about which is preferable: fixing this instance of #139 directly or downgrading to an error message...

stweil:
Issue #139 is fixed now with a better solution, so closing this pull request.

PR #144, opened 2020-08-24
Segmentation on raw images

bertsky:


bertsky:
This needs to be tested systematically. I expect to see **both** degradation and improvement, depending on how hard binarization is. See [here](https://github.com/tesseract-ocr/tesseract/issues/3083) for explanation.

kba (commented):
I understand the reasoning, subscribing to https://github.com/tesseract-ocr/tesseract/issues/3083 for the discussion on upstream changes. Changeset (filtering `binarized`) is sensible but needs good testing to ensure that it is more beneficial than detrimental, or perhaps should be parameterizable.

bertsky:
> or perhaps should be parameterizable.

I thought about that, but at workflow configuration time, you have next to no chance of knowing which is going to be better. (I would guess that only input images which fare well under global Otsu are better off with the change. But we have no automatic indicator of binarization quality yet. In the very least, we should strive for some estimator based on local distribution of connected component statistics.)

But I still hope that we can fix the problem in Tesseract itself.

PR #158, opened 2020-10-19
Add fontshape processor and all-in-one segmentation

bertsky:
We can probably remove **both** the old `segment-region/line/word` and new (all-in-one) `segment` altogether now that we can configure them via `overwrite_*` and `textequiv_level` in `recognize`. Or we keep the CLI names, but delegate to `recognize` @kba?


ocrd_tesserocr/fontshape.py
```diff
@@ -0,0 +1,189 @@
...
+                    if 'bold' in word_attributes else None,
+                    italic=word_attributes['italic']
+                    if 'italic' in word_attributes else None,
+                    underlined=word_attributes['underlined']
+                    if 'underlined' in word_attributes else None,
+                    monospace=word_attributes['monospace']
+                    if 'monospace' in word_attributes else None,
+                    serif=word_attributes['serif']
+                    if 'serif' in word_attributes else None)
+                word.set_TextStyle(word_style) # (or somewhere in custom attribute?)
```
kba, line 175: Excellent.

I think `<TextStyle>` is the right place, c.f. https://ocr-d.de/en/spec/page#textstyle

ocrd_tesserocr/recognize.py
```diff
@@ -56,26 +67,65 @@ def __init__(self, *args, **kwargs):
...
+        - If ``overwrite_words``, segment words, else iterate existing words.
+        - If ``textequiv_level==word``, then recognize text in the words,
+          annotate it, and continue with the next word. Otherwise...
+        - If ``textequiv_level==glyph``, then recognize text in the glyphs and
+          continue with the next glyph. Otherwise...
+        - (i.e. ``none``) annotate no text and be done with segmentation.
+        
+        Thus, enabling the ``overwrite_*`` modes makes this processor behave more
+        like a segmentation processor, and setting ``textequiv_level`` to ``none``
+        makes it a segmentation-only processor.
```
kba, line 116: OK, just so I understand this, `recognize -P overwrite_lines -P textequiv_level none` is equivalent to `segment-lines` now, correct?

setup.py
```diff
@@ -1,15 +1,17 @@
 # -*- coding: utf-8 -*-
 """
-Installs five executables:
+Installs the following command-line executables:
 
+    - ocrd_tesserocr_binarize
```
kba, line 5: ocrd-tesserocr-binarize etc. (dashes not underscores)

kba (approved):
Great stuff

ocrd_tesserocr/recognize.py
kba: yes, see comment on PR :tada:

ocrd_tesserocr/fontshape.py
bertsky: Yes, but maybe we should also "propagate" it to the parents' `TextStyle` analogously to `TextEquiv`.

The problem with word-wise recognition is that it's not as reliable. But if we allow the pre-LSTM models to operate on the complete line, they will (in general) come up with a different segmentation than the one annotated in PAGE from some LSTM recognizer.

One improvement I had in mind was cropping word images from line y and word x, so that words without ascenders or descenders (like `von`) are not considered larger than others merely because they are cropped tighter.

ocrd_tesserocr/recognize.py
bertsky: Yes, exactly!

And if you add `-P overwrite_regions true` you get the `ocrd-cis-ocropy-segment` and `sbb-textline-detector` behaviour.

kba:
> We can probably remove **both** the old `segment-region/line/word` and new (all-in-one) `segment` altogether now that we can configure them via `overwrite_*` and `textequiv_level` in `recognize`. Or we keep the CLI names, but delegate to `recognize` @kba?

We should keep the `segment{,-{region,word,line}` CLIs so we don't break existing workflows. If they can delegate to `recognize` with fixed `textequiv_level` and `overwrite_LEVEL` parameters, that would be best. And reduce code duplication, making recognize more complex but much more powerful.

setup.py
bertsky: Well spotted!

bertsky:
>  If they can delegate to `recognize` with fixed `textequiv_level` and `overwrite_LEVEL` parameters, that would be best.

One problem I still see with such delegation is that we would still need to repeat the tool json. And we would still have to write some thin module instantiating `TesserocrRecognize` and translating the call.

kba:
> One problem I still see with such delegation is that we would still need to repeat the tool json. And we would still have to write some thin module instantiating TesserocrRecognize and translating the call.

Duplicate info in the ocrd-tool.json is okay I think, it's one place and the user doesn't normally see it. If that is too error-prone, a generator script for ocrd-tool.json might help.

> thin module instantiating TesserocrRecognize and translating the call.

Yes, but you could inherit/encapsulate `TesserOcrRecognize` and do smth like

```python
class TesserocrLineSegmenter(TesserocrRecognize):
  def process(self):
    self.parameter['overwrite_lines'] = True
    self.parameter['textequiv_level'] = "none"
    return super().process(self)
```

bertsky:
> Yes, but you could inherit/encapsulate `TesserOcrRecognize` and do smth like
> 
> ```python
> class TesserocrLineSegmenter(TesserocrRecognize):
>   def process(self):
>     self.parameter['overwrite_lines'] = True
>     self.parameter['textequiv_level'] = "none"
>     return super().process(self)
> ```

It's not quite as easy I'm afraid. In the constructor, we must pick the derived tool from the tool.json `tools` (or `-h` and `-J` won't work.) So we need to parse the derived tool json and apply it against the input params, but then in `process` have to fill in the defaults of the original tool json. So the superclass' json must be dumped, a `ParameterValidator` be constructed for that, and applied against the input params a second time. 

Or am I getting it all wrong?

kba:
How about this:

```python
from ocrd import Processor
from .config import OCRD_TOOL

from .recognize import TesserocrRecognize
TOOL = 'ocrd-tesserocr-segment-line'

class TesserocrSegmentLine(Processor):

        kwargs['version'] = OCRD_TOOL['version']
        super().__init__(*args, **kwargs)

        recognize_kwargs = {**kwargs}
        recognize_kwargs.pop('show_help', None)
        recognize_kwargs.pop('show_version', None)
        recognize_kwargs['parameter'] = {}
        recognize_kwargs['parameter']['overwrite_lines'] = True
        recognize_kwargs['parameter']['textequiv_level'] = "none"
        self.recognizer = TesserocrRecognize(**recognize_kwargs)

    def process(self):
        return self.recognizer.process()
```

bertsky:
> How about this:

We still need to do sth about the workspace argument. Maybe pass `None` so it also won't chdir a second time?

kba:
> > How about this:
> 
> We still need to do sth about the workspace argument. Maybe pass `None` so it also won't chdir a second time?

You mean "... and assign `workspace` to `self.recognizer` manually after instantiation"?

i.e.
```python
        recognize_kwargs.pop('show_help', None)
        self.recognizer = TesserocrRecognize(**recognize_kwargs)
        self.recognizer.workspace = self.workspace
```

That would work if the second chdir bothers you.

bertsky:
> That would work if the second chdir bothers you.

You still need to pass `None` for instantiation (workspace is not a kwarg, and chdir would fail).

> ```python
>         self.recognizer.workspace = self.workspace
> ```

Yes, in combination with that it should work.

bertsky:
Trying to make the `segment*` CLIs delegate to `recognize`, I realize that we do still have a conceptual discrepancy: The latter is not capable of incremental annotation, so it does not treat the `overwrite_*` parameters as the former do. (Even the former were not truly incremental, they just added new segments redundantly. A true incremental behaviour would mask the existing segments.) How should we approach this?

bertsky:
> Trying to make the `segment*` CLIs delegate to `recognize`, I realize that we do still have a conceptual discrepancy: The latter is not capable of incremental annotation, so it does not treat the `overwrite_*` parameters as the former do. (Even the former were not truly incremental, they just added new segments redundantly. A true incremental behaviour would mask the existing segments.) How should we approach this?

Here's my proposal: Since the existing `overwrite_*` parameters did not really do anything useful (as they merely allowed redundant segments when `False`), we just ignore them from now on. The new `overwrite_*` parameters in `recognize` are independent. If I add some form of incremental annotation some day, I will slightly change their meaning (still triggering segmentation, but not removing existing segments/areas anymore).

kba:
> we just ignore them from now on

As far as I can grok the consequences, that seems reasonable.

bertsky:
> The new `overwrite_*` parameters in `recognize` are independent. If I add some form of incremental annotation some day, I will slightly change their meaning (still triggering segmentation, but not removing existing segments/areas anymore).

Or, we might already give them **different** names in `recognize`, say `detect_regions`, `detect_lines`, `detect_words` (and have no `overwrite_*` for now). That way, once we do add incremental annotation, we can still turn it off via `overwrite_regions=True` etc.

Ignoring the old `overwrite_*` would be as simple as suppressing them in the parameter dict during delegation.

kba:
> > The new `overwrite_*` parameters in `recognize` are independent. If I add some form of incremental annotation some day, I will slightly change their meaning (still triggering segmentation, but not removing existing segments/areas anymore).
> 
> Or, we might already give them **different** names in `recognize`, say `detect_regions`, `detect_lines`, `detect_words` (and have no `overwrite_*` for now). That way, once we do add incremental annotation, we can still turn it off via `overwrite_regions=True` etc.

`detect_LEVEL` is certainly more intuitive than `overwrite_LEVEL` with a different meaning in different contexts.

bertsky:
> `detect_LEVEL` is certainly more intuitive than `overwrite_LEVEL` with a different meaning in different contexts.

Agreed. Last one: `detect_regions` or `segment_regions`?

kba:
I think `detect_LEVEL` is better than `segment_LEVEL` because in the former it is clear that LEVEL is what is being detected, whereas in the latter, it could mean "find LEVEL+1 in LEVEL".

bertsky:
> Ignoring the old `overwrite_*` would be as simple as suppressing them in the parameter dict during delegation.

Since this part of the discussion was all about backwards-compatibility, I think it would be fair if we made these parameters "true-only" in the tool json (via `"enum": [true]`). That way, if anyone used `False` in their workflow, they will know they have to look at this and change their configuration.

kba:
> > That would work if the second chdir bothers you.
> 
> You still need to pass `None` for instantiation (workspace is not a kwarg, and chdir would fail).

Sure, copy/paste 
> 
> > ```python
> >         self.recognizer.workspace = self.workspace
> > ```
> 
> Yes, in combination with that it should work.



> > Ignoring the old `overwrite_*` would be as simple as suppressing them in the parameter dict during delegation.
> 
> Since this part of the discussion was all about backwards-compatibility, I think it would be fair if we made these parameters "true-only" in the tool json (via `"enum": [true]`). That way, if anyone used `False` in their workflow, they will know they have to look at this and change their configuration.

:+1: I've never used enums for anything other than strings acoording to https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values `{"type": boolean, "enum": [true]}` is indeed a valid construct. Nice, learned something new.

bertsky:
>  I've never used enums for anything other than strings acoording to https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values `{"type": boolean, "enum": [true]}` is indeed a valid construct. Nice, learned something new.

Yes, and it will give these hilarious tautological responses:
```
Invalid parameters ['[overwrite_words] False is not one of [True]']
```

ocrd_tesserocr/fontshape.py
bertsky: Tried this, and gave up: It does not help getting more homogeneous font sizes, but increases the likelihood of having intruding components from neighouring lines in the word before recognition (which also impairs font attribute quality).

bertsky:
Sorry, when trying to add the table processing options in a reasonable way (with an extra `cell` level in between) I found that the most sensible parameterization would be `segmentation_level` (instead of `detect_*`) analogous to `textequiv_level`. So we finally came back to your [input-level/output-level](https://github.com/OCR-D/spec/issues/134) idea :-)

It's not _really_ input vs output level of course. See extensive documentation in docstring and tool json.

Compared to processors with a `level-of-operation`, the semantics is one step below here (which simply makes more sense facing Tesseract's API).

Delegation works perfectly BTW. (I just replaced the logger name. The PAGE metadata will show the true processor's name and parameters though, which is ok I guess.)

kba (approved):
> segmentation_level (instead of detect_*) analogous to textequiv_level

Also a consistent naming scheme, fine by me.

> Delegation works perfectly BTW.

Looks good. It will do the chdir twice like this though.

> The PAGE metadata will show the true processor's name and parameters though, which is ok I guess.

Yeah, makes sense to have the parameterization of the actual processor in there.

bertsky:
Just saw (in time before merging) that I forgot to keep `overwrite_words` in recognize for backwards compatibility.

I think I'll re-add this and if
- `true` (probably in some existing workflows): verify that `segmentation_level=word` at the same time (with an exception)
- `false` (the old default): verify that _not_ `segmentation_level=word` at the same time (with an exception)

(I could also try as `type: object` and `enum: [false, true, "none"]` with `none` as new default and allowing to forget about this parameter and its interaction with `segmentation_level` in the future. But then again, it's probably just too complicated.)

kba:
> Just saw (in time before merging) that I forgot to keep `overwrite_words` in recognize for backwards compatibility.
> 
> I think I'll re-add this and if
> 
>     * `true` (probably in some existing workflows): verify that `segmentation_level=word` at the same time (with an exception)
> 
>     * `false` (the old default): verify that **not** `segmentation_level=word` at the same time (with an exception)

 
> (I could also try as `type: object` and `enum: [false, true, "none"]` with `none` as new default and allowing to forget about this parameter and its interaction with `segmentation_level` in the future. But then again, it's probably just too complicated.)

I actually think that's a good solution, users would see these three values as "Possible values" in the help and it would make it clear, that the parameter is deprecated and should be left at `none` for future invocations.

bertsky:
> > Just saw (in time before merging) that I forgot to keep `overwrite_words` in recognize for backwards compatibility.
> > I think I'll re-add this and if
> > ```
> > * `true` (probably in some existing workflows): verify that `segmentation_level=word` at the same time (with an exception)
> > 
> > * `false` (the old default): verify that **not** `segmentation_level=word` at the same time (with an exception)
> > ```
> 
> > (I could also try as `type: object` and `enum: [false, true, "none"]` with `none` as new default and allowing to forget about this parameter and its interaction with `segmentation_level` in the future. But then again, it's probably just too complicated.)
> 
> I actually think that's a good solution, users would see these three values as "Possible values" in the help and it would make it clear, that the parameter is deprecated and should be left at `none` for future invocations.

Thanks for that assessment. Follow-up: So when I do have true/false, not `none` or `ignore`, then I must check against `segmentation_level`. But in my `process()`, it's too late knowing whether the user explicitly requested the latter, or its value came in from the default. So the question becomes: Do I check `segmentation_level` in the kwarg parameters (before calling the superclass' constructor) or the resulting state (as documented with default value) to decide on raising an exception or not?

Sorry to be so obsessed with details here, but I think it's worth the effort, because we want to inflict as little pain and wonder as possible for existing workflows/users.

bertsky:
> So the question becomes: Do I check `segmentation_level` in the kwarg parameters (before calling the superclass' constructor) or the resulting state (as documented with default value) to decide on raising an exception or not?

In practice, for existing workflows with `overwrite_words=true` the latter will raise an exception, whereas the former will silently set `segmentation_level=word`. I am inclined towards the latter, also because we can raise more understandable exceptions where we have to.

bertsky:
> > So the question becomes: Do I check `segmentation_level` in the kwarg parameters (before calling the superclass' constructor) or the resulting state (as documented with default value) to decide on raising an exception or not?
> 
> In practice, for existing workflows with `overwrite_words=true` the latter will raise an exception, whereas the former will silently set `segmentation_level=word`. I am inclined towards the latter, also because we can raise more understandable exceptions where we have to.

On the other hand, this behaviour might be even more unexpected. Maybe I should just be bold and set `segmentation_level=word` – no exception, just a warning (and a plea to update the workflow)?

bertsky:
> > > So the question becomes: Do I check `segmentation_level` in the kwarg parameters (before calling the superclass' constructor) or the resulting state (as documented with default value) to decide on raising an exception or not?
> > 
> > 
> > In practice, for existing workflows with `overwrite_words=true` the latter will raise an exception, whereas the former will silently set `segmentation_level=word`. I am inclined towards the latter, also because we can raise more understandable exceptions where we have to.
> 
> On the other hand, this behaviour might be even more unexpected. Maybe I should just be bold and set `segmentation_level=word` – no exception, just a warning (and a plea to update the workflow)?

```
'ignore' is not of type 'object'
```

Should've known better. This 3-valued-boolean, have-my-cake-and-eat-it-too compatibility thing was a mistake. I'm going to make this a ignore-and-warn-only thing altogether.

bertsky:
> > > > So the question becomes: Do I check `segmentation_level` in the kwarg parameters (before calling the superclass' constructor) or the resulting state (as documented with default value) to decide on raising an exception or not?
> > > 
> > > 
> > > In practice, for existing workflows with `overwrite_words=true` the latter will raise an exception, whereas the former will silently set `segmentation_level=word`. I am inclined towards the latter, also because we can raise more understandable exceptions where we have to.
> > 
> > 
> > On the other hand, this behaviour might be even more unexpected. Maybe I should just be bold and set `segmentation_level=word` – no exception, just a warning (and a plea to update the workflow)?
> 
> ```
> 'ignore' is not of type 'object'
> ```
> 
> Should've known better. This 3-valued-boolean, have-my-cake-and-eat-it-too compatibility thing was a mistake. I'm going to make this a ignore-and-warn-only thing altogether.

Reverted the whole thing and made the hard choice of _not_ being backwards compatible at all here. If there's any time when users must pay attention and rewrite their configs, it's now.

stweil:
[My ocrd process](https://github.com/UB-Mannheim/ocrd_all/wiki/urn:nbn:de:bsz:180-digad-22977#workflow-2) fails to recognize text with this pull request.

bertsky:
> [My ocrd process](https://github.com/UB-Mannheim/ocrd_all/wiki/urn:nbn:de:bsz:180-digad-22977#workflow-2) fails to recognize text with this pull request. I tested [40de199](https://github.com/OCR-D/ocrd_tesserocr/commit/40de199d844b1dcad1a0f3743ddca312169e904e) and [24b7ced](https://github.com/OCR-D/ocrd_tesserocr/commit/24b7ced7379afab821020c49f857cc79bc949fbc), both failed.

40de199 is an early version of this PR, without re-parameterization and all the backwards compatibility considerations. You should use the latest commit to assess this PR.

24b7ced is the current master and v0.9.5. If that failed already, then it seems your problem is not related to this PR, and I suggest you open a new issue with a full description of what went wrong (including logs).

BTW your workflow configuration has a bug: you cannot use clipping after a processor like deskewing which creates derived images on the same level. I am getting really frustrated about this now. I keep telling people this __does not work__, the log messages tell you the same, the help text as well. I have no idea how this ever came to be part of our workflow guide, which [along with other bad recommendations](https://github.com/OCR-D/ocrd-website/issues/172) is still up there on ocr-d.de.

stweil:
Sorry @bertsky for my wrong information (maybe I should stop working for today). I had first tested with an older version of your PR and @kba's PR for ocrd_all. That failed. Then I noticed that on another computer @kba's PR alone worked fine. I then tested the current bertsky/add-word-font-attributes, and it failed, too. I know that the workflow does not give the best results, but I used it because it is one of the "recommended" OCR-D workflows. And I still use it to compare the results with previous runs.

bertsky:
> Sorry @bertsky for my wrong information (maybe I should stop working for today). I had first tested with an older version of your PR and @kba's PR for ocrd_all. That failed. Then I noticed that on another computer @kba's PR alone worked fine. I then tested the current bertsky/add-word-font-attributes, and it failed, too. I know that the workflow does not give the best results, but I used it because it is one of the "recommended" OCR-D workflows. And I still use it to compare the results with previous runs.

Understood. I can see that it is hard to move away from a workflow once you have established it as a baseline, without regression checking that is. But at least the `clip` in there should be a (computationally expensive) no-op.

I would like to merge this PR. Could you please give it a try again?

bertsky:
@stweil you stated that this (an early version of) PR broke recognition for you. Can you please revisit, so I can merge? 

In [your workflow](https://github.com/UB-Mannheim/ocrd_all/wiki/urn:nbn:de:bsz:180-digad-22977#workflow-2), you can simplify the steps between `OCR-D-BIN-DENOISE-DESKEW` and `OCR-D-OCR-TESS` like so:

    "tesserocr-segment -I OCR-D-BIN-DENOISE-DESKEW -O OCR-D-SEG-LINE" \
    "tesserocr-deskew -I OCR-D-SEG-LINE -O OCR-D-SEG-LINE-FLIP -P operation_level region" \
    "cis-ocropy-deskew -I OCR-D-SEG-LINE-FLIP -O OCR-D-SEG-LINE-DESKEW -P level-of-operation region" \
    "tesserocr-recognize -I OCR-D-SEG-LINE-DESKEW -O OCR-D-OCR-TESS -P model Fraktur_50000000.334_450937" \

stweil:
I just tested 9118efd989126ed75352caac3dc0345603964806 (which only differs in `README.md` from your latest commit). It is still broken: `ocrd-tesserocr-recognize -I OCR-D-SEG-LINE -O OCR-D-OCR-TESS2 -P model fast/Fraktur_50000000.502_198857` did not recognize usable text.

bertsky:
> I just tested [9118efd](https://github.com/OCR-D/ocrd_tesserocr/commit/9118efd989126ed75352caac3dc0345603964806) (which only differs in `README.md` from your latest commit). It is still broken: `ocrd-tesserocr-recognize -I OCR-D-SEG-LINE -O OCR-D-OCR-TESS2 -P model fast/Fraktur_50000000.502_198857` did not recognize usable text.

I cannot reproduce this kind of problem. Could you please share your METS, or make a minimal example (just one page's original image + line segmented PAGE)?

> Should I re-run a test with the suggested simplifications?

It's up to you (would be good to have anyway). I'll first try to isolate the issue you see.

stweil:
Here is the complete data set of the build: https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201023/.

bertsky:
> Here is the complete data set of the build: https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201023/.

Thx. The log says it came to start deskewing (but didn't complete it). But you were saying that `recognize` did not produce usable text (which comes with the presupposition that it did at least start and did produce _some_ text). So is this workspace just in a different state than what you based your report on? Also, I'd like to download this, but the server says robots aren't allowed... (The next best thing I could try is download your METS only, but then I'd have to try to reconstruct all the remote links in there.)

stweil:
Please read my updated comment. The log which I first linked is from the still running new OCR-D process.

stweil:
@bertsky, the modified OCR-D process aborted with `AttributeError: 'NoneType' object has no attribute 'getbbox'`, so there is another issue with your pull request. See https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201023/ocr.log for details.

stweil (changes requested):
See conversation.

bertsky:
> Please read my updated comment. The log which I first linked is from the still running new OCR-D process.

Ah, got it. Unfortunately, https://github.com/OCR-D/core/issues/403 casts its shadow – I get a 403 on the mets.xml :roll_eyes:

Also, because of https://github.com/OCR-D/core/issues/592 your log does not contain any hints on what happened.

Anyway, I managed to reconstruct parts of your workspace via `bulk-add` and can **reproduce** the problem :tada:

The log is filled with messages like this:
```
WARNING processor.TesserocrRecognize - Line 'region0003_line0000' contains no words (but segmentation if off)
```

That quickly helped me identify the problem: the default for `segmentation_level` (for backwards compatibility and under the principle of least astonishment) needs to be `word`, not `none`. 

Wow, I wonder why I did not detect this myself. Must be the wood from the trees. :confused:

Anyway, thanks @stweil – that change was crucial!

Meanwhile, I also added orientation and skew detection to `recognize` (i.e. those parts of the dedicated `deskew` processor that don't need loading the special model `osd` in a legacy OEM mode). This should be without consequence for the all-in-one use case (all segmentation and recognition in one step from the same iterator), because Tesseract rotates itself internally, but in all other cases it automatically makes the subsequent processors rotate accordingly (without the need for an extra deskewing step in the workflow). 

So my recommendation…

> "tesserocr-segment -I OCR-D-BIN-DENOISE-DESKEW -O OCR-D-SEG-LINE" \
> "tesserocr-deskew -I OCR-D-SEG-LINE -O OCR-D-SEG-LINE-FLIP -P operation_level region" \
> "cis-ocropy-deskew -I OCR-D-SEG-LINE-FLIP -O OCR-D-SEG-LINE-DESKEW -P level-of-operation region" \
> "tesserocr-recognize -I OCR-D-SEG-LINE-DESKEW -O OCR-D-OCR-TESS -P model Fraktur_50000000.334_450937" \

…now became…

> "tesserocr-segment -I OCR-D-BIN-DENOISE-DESKEW -O OCR-D-SEG-LINE" \
> "cis-ocropy-deskew -I OCR-D-SEG-LINE -O OCR-D-SEG-LINE-DESKEW -P level-of-operation region" \
> "tesserocr-recognize -I OCR-D-SEG-LINE-DESKEW -O OCR-D-OCR-TESS -P model Fraktur_50000000.334_450937" \

…but could of course be simplified to just…

> "tesserocr-recognize -I OCR-D-BIN-DENOISE-DESKEW -O OCR-D-OCR-TESS -P segmentation_level region -P model Fraktur_50000000.334_450937" \

…if the (usually) increased deskewing accuracy of Ocropy is not needed.

bertsky:
@stweil 
> the modified OCR-D process aborted with `AttributeError: 'NoneType' object has no attribute 'getbbox'`, so there is another issue with your pull request

Fixed that, too. (Tesseract seems to have different kinds of empty regions.)

stweil:
The latest code works again, but gives different OCR results in my test workflow. https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201022/urn:nbn:de:bsz:180-digad-22977/ contains the results with `OCR-D-OCR-TESS` (old results), `OCR-D-OCR-TESS2` (broken PR, only first few pages), `OCR-D-OCR-TESS3` (latest PR, only first few pages).

I also tested the simplified workflow as suggested. It now works, too, with slightly worse OCR results: https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201024/urn:nbn:de:bsz:180-digad-22977/

bertsky:
> The latest code works again, but gives different OCR results in my test workflow.

On a glance, subjectively, they actually look better IMO. But could you please re-run with `-P padding 0` just to make sure the difference can be attributed to that new parameter default? 

> I also tested the simplified workflow as suggested. It now works, too, with slightly worse OCR results

When you say slightly worse, how are you measuring – against some GT via concatenation+alignment+CER-calculation? Or just random sampling with intellectual assessment? Can you point me to a concrete (ideally typical) example of deterioration?

stweil:
With padding 0 I get the old result, so yes, the difference was caused by that new default value.

I only had a short look on the deltas, so when I said the new results are slightly worse this is subjective. Here is an example which compares old and new results:

    -            <pc:TextEquiv conf="0.875303970402969">
    -                <pc:Unicode>M Qlerliebſẽ in Chriſto bi ba be
    +            <pc:TextEquiv conf="0.86770378157862">
    +                <pc:Unicode>— Qerlebnt in Chꝛiſt⸗ bi ba bt
     jr die Chriſtliche anzaigung in die haylige
    -geſchrifft gottes / vnſers lieben bruͤdern Philips
    -Melanchthon / wie er ſe vnns allen zů haylſa⸗
    -mer vnderrichtung / Fuͤm andern mal mitt etli⸗
    -cher aͤnderũg im truck hat laſſen außgoͤn / durch
    +geſchrifft gottes/ vnſers lieben bruͤdern Philips
    +Melanchthon / wie er ſie vnns allen zů haylſa⸗
    +mer vnderrichtung / Zuͤm andern mal mitt etli⸗
    +cher aͤnderũg im truckhat laſſen außgoͤn / durch
     mich verteütſcht. Derhalbẽ iſt an eüch alle mein
    -fleyſſig bitt / jr woͤllet diß haylwaͤrtigs buͤchlin
    +fleyſſig bitt / jr woͤllet diß haylwaͤrtigs buchlin
     treülich fleiſſig vnd offt laͤſen. Angeſehen das es
     vns den richtigen geraden weg zů dem gebene⸗
     deyten raynenn wort Gottes zaiget. Das S.
    -Panl nẽnet die krafft gottes dardurch ſaͤlig wer⸗
    +Paul nẽnet die krafft gottes dardurch ſaͤlig wer⸗
     den alle die daran glauben / das wort der war⸗
     hait / das hayl gottes / vnd das Euangelium das
     iſt die gůt botſchafft vnſer ſaͤligkait. Dartzů got
     der allmaͤchtig vatter durch ſein ainigen ſon Je⸗
    -ein e err ríν-
    +n neerríνν-—
     vns allen mit genaden haͤlffe mit aynem ſaͤligen
    -zuů nemen vnd wachſen in der genaden vnnd er⸗
    -kaͤntnüß des herren Iheſu Chriſti/ Amen. Da⸗
    +zů nemen vnd wachſen in der genaden vnnd er⸗
    +kaͤntnüß des herren Iheſu Chriſti / Amen. Da⸗
     tum am vierdẽ tag des Octobers. Anno domini.
    -M. D. xvxij.</pc:Unicode>
    +M. D. xxij.</pc:Unicode>
                 </pc:TextEquiv>

In this example there is no clear winner. Sometimes old is better, sometimes new is better.

stweil:
When comparing the old workflow with no padding to the simplified new workflow, the deltas are clearer:

    -            <pc:TextEquiv conf="0.875303970402969">
    -                <pc:Unicode>M Qlerliebſẽ in Chriſto bi ba be
    +            <pc:TextEquiv conf="0.830030305528709">
    +                <pc:Unicode>H eledſen Chriſto bi bat
     jr die Chriſtliche anzaigung in die haylige
    -geſchrifft gottes / vnſers lieben bruͤdern Philips
    -Melanchthon / wie er ſe vnns allen zů haylſa⸗
    -mer vnderrichtung / Fuͤm andern mal mitt etli⸗
    +geſchrifft gottes/ vnſers lieben bruͤdern Philips
    +clanchchon / wie er ße vnns allen ʒů haylſa⸗
    +mer vnderrichtun g / Fuͤm an dern mal mitt etli⸗
     cher aͤnderũg im truck hat laſſen außgoͤn / durch
    -mich verteütſcht. Derhalbẽ iſt an eüch alle mein
    -fleyſſig bitt / jr woͤllet diß haylwaͤrtigs buͤchlin
    -treülich fleiſſig vnd offt laͤſen. Angeſehen das es
    +mich verteücſcht. Derhalbẽ iſt an eüch alle mein
    +fleyſſig bitt / jr woͤllet diß haylwaͤrcigs bůchlin
    +treülich fleiſſig vnd offt laͤſen . Angeſehen das es
     vns den richtigen geraden weg zů dem gebene⸗
     deyten raynenn wort Gottes zaiget. Das S.
    -Panl nẽnet die krafft gottes dardurch ſaͤlig wer⸗
    +Haul nẽnet die krafft gottes dardurch ſaͤlig wer⸗
     den alle die daran glauben / das wort der war⸗
    -hait / das hayl gottes / vnd das Euangelium das
    -iſt die gůt botſchafft vnſer ſaͤligkait. Dartzů got
    +hait /das hayl gottes / vnd das Euangelium das
    +iſt die gůt botſchafft vnſer faͤligkait. Dartzu got
     der allmaͤchtig vatter durch ſein ainigen ſon Je⸗
    -ein e err ríν-
    +—
     vns allen mit genaden haͤlffe mit aynem ſaͤligen
    -zuů nemen vnd wachſen in der genaden vnnd er⸗
    +zů nemen vnd wachſen in der genaden vnnd er⸗
     kaͤntnüß des herren Iheſu Chriſti/ Amen. Da⸗
     tum am vierdẽ tag des Octobers. Anno domini.
    -M. D. xvxij.</pc:Unicode>
    +M. D. xxij.</pc:Unicode>
                 </pc:TextEquiv>

stweil (approved):
The latest code works in my tests. It gives slightly different OCR results with the new default for padding, so please consider using the old default until there is clear evidence that the changed default is superior.

bertsky:
> Here is an example which compares old (no padding) and new (padding) results:
[...]
> In this example there is no clear winner. Sometimes old is better, sometimes new is better.

> When comparing the old workflow with no padding to the simplified new workflow, the deltas are clearer:
[...]

Yes. Maybe to get a clearer impression we should look at the images that Tesseract is presented with, or which it crops internally.

And really I think we should evaluate this on a larger sample automatically via CER. Maybe OCR-D text GT is suited best for this.

The reasoning for the new default came from (other other subjective measurements and) the internal Tesseract behaviour. Before LSTM decoding, the image is [padded](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccmain/linerec.cpp#L244) by [4px](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccstruct/imagedata.h#L38-L39). (This seems to be a requirement from the LSTM modelling or training.)

But this does not always have an effect, as the [padded bbox](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccmain/linerec.cpp#L172) then gets [intersected](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccmain/linerec.cpp#L192-L194) with the image frame. (No white pixels are added around it.)

So, when we pass an image in `PSM.SINGLE_WORD` (as in `ocrd-tesserocr-recognize` with `segmentation_level=glyph` or `textequiv_level=word`) or `PSM.RAW_LINE` (as in `ocrd-tesserocr-recognize` with `segmentation_level=word` or `textequiv_level=line`, and `raw_lines=true`), then _actually_ nothing is padded. And in `PSM.SINGLE_LINE` (as in `ocrd-tesserocr-recognize` with `segmentation_level=word` or `textequiv_level=line`, but `raw_lines=false`) there is _some_ padding, but only up to the margin that the internal single-line segmentation of the image detected.

On the other hand, when entering from higher iterator positions (above the line level, i.e. line or region+line segmentation with recognition in a single step), then (except for the top left and bottom right of the image) said 4px internal padding _will_ apply to each line.

Thus, when the segmentation is external to `ocrd-tesserocr-recognize`, padding needs to be applied externally, too. (This is what the `pad_image` function does in recognize.py.) And when the segmentation is internal, but may be exported via PAGE and then re-imported by later processors, the same kind of padding should be added explicitly to the bbox. (This is what the `padding` kwarg to the Tesseract `PageIterator.BoundingBox` method is doing in recognize.py.)

bertsky:
> Yes. Maybe to get a clearer impression we should look at the images that Tesseract is presented with, or which it crops internally.

It does seem to be related to chopped off intruders from neighbouring components at the margins. When you pad beyond them, they become independent features, which naturally are more confusing than too tightly cropped main features. But this is just a first impression.

> And really I think we should evaluate this on a larger sample automatically via CER. Maybe OCR-D text GT is suited best for this.

This will take longer, so I suggest tracking this in an extra issue or PR.

> The reasoning for the new default came from (other subjective measurements and) the internal Tesseract behaviour: Before LSTM decoding, the image is [padded](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccmain/linerec.cpp#L244) by [4px](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccstruct/imagedata.h#L38-L39). (This seems to be a requirement from the LSTM modelling or training.)
> 
> But this does not always have an effect, as the [padded bbox](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccmain/linerec.cpp#L172) then gets [intersected](https://github.com/tesseract-ocr/tesseract/blob/11297c983ec7f5c9765d7fa4faa48f5150cf2d38/src/ccmain/linerec.cpp#L192-L194) with the image frame. (No white pixels are added around it.)
> 
> So, when we pass an image in `PSM.SINGLE_WORD` (as in `ocrd-tesserocr-recognize` with `segmentation_level=glyph` or `textequiv_level=word`) or `PSM.RAW_LINE` (as in `ocrd-tesserocr-recognize` with `segmentation_level=word` or `textequiv_level=line`, and `raw_lines=true`), then _actually_ nothing is padded. And in `PSM.SINGLE_LINE` (as in `ocrd-tesserocr-recognize` with `segmentation_level=word` or `textequiv_level=line`, but `raw_lines=false`) there is _some_ padding, but only up to the margin that the internal single-line segmentation of the image detected.
> 
> On the other hand, when entering from higher iterator positions (above the line level, i.e. line or region+line segmentation with recognition in a single step), then (except for the top left and bottom right of the image) said 4px internal padding _will_ apply to each line.
> 
> Thus, when the segmentation is external to `ocrd-tesserocr-recognize`, padding needs to be applied externally, too. (This is what the `pad_image` function does in recognize.py.) And when the segmentation is internal, but gets exported via PAGE and may then be re-imported by later processors, the same kind of padding should be added explicitly to the bbox. (This is what the `padding` kwarg to the Tesseract `PageIterator.BoundingBox` method is doing in recognize.py.)

Maybe I should split the bbox/output padding (from higher iterator positions) from the image/input padding (at the entry level), as the latter is new to this processor and might be more dangerous? I see a number of options how to proceed here:

1. keep image padding, but set defaults for both kinds of padding to 0px (as before)
1. remove image padding, start a new PR for it (and get involved in the GT measurements there)
1. expose a different parameter name for image padding, say `image_padding`, which defaults to 0px

wrznr:
As stated in personal communication, I'd opt for 1 because it ensures consistency with existing workflows and allows to proceed.

bertsky:
@wrznr 
> As stated in personal communication, I'd opt for 1 because it ensures consistency with existing workflows and allows to proceed.

But you also recommended against backwards compatibility earlier (regarding the `overwrite_words` / `segmentation_level` re-parameterization). Users will have to adapt their workflows at this point, anyway. I think that we should use that chance to change defaults more sensibly. When I saw that input padding can actually hurt, I agreed to postpone this, but output padding really is a different thing. I am strongly inclined towards 2...

wrznr:
AFAIR, I made the point that we should not let ourself restrict or be dominated by the backwards compatibility principle. But if you **strongly** tend towards 2, you of course have my placet to proceed.

stweil:
I just have run a new test with latest ocrd_all + latest version of this pull request and see a major regression with many lines of text missing in the final result (compare [old](https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201022/urn:nbn:de:bsz:180-digad-22977/OCR-D-OCR-TEXT/FILE_0007_OCR-D-OCR-TEXT.txt) with [new](https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201105/urn:nbn:de:bsz:180-digad-22977/OCR-D-OCR-TEXT/FILE_0007_OCR-D-OCR-TEXT.txt)).

bertsky:
> I just have run a new test with latest ocrd_all + latest version of this pull request and see a major regression with many lines of text missing in the final result (compare [old](https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201022/urn:nbn:de:bsz:180-digad-22977/OCR-D-OCR-TEXT/FILE_0007_OCR-D-OCR-TEXT.txt) with [new](https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201105/urn:nbn:de:bsz:180-digad-22977/OCR-D-OCR-TEXT/FILE_0007_OCR-D-OCR-TEXT.txt)).

Thanks @stweil! I forgot I need to jump across `RIL_PARA` when I want text lines in text blocks.

kba:
> > I just have run a new test with latest ocrd_all + latest version of this pull request and see a major regression with many lines of text missing in the final result (compare [old](https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201022/urn:nbn:de:bsz:180-digad-22977/OCR-D-OCR-TEXT/FILE_0007_OCR-D-OCR-TEXT.txt) with [new](https://ub-backup.bib.uni-mannheim.de/~stweil/OCR-D/pilot/20201105/urn:nbn:de:bsz:180-digad-22977/OCR-D-OCR-TEXT/FILE_0007_OCR-D-OCR-TEXT.txt)).
> 
> Thanks @stweil! I forgot I need to jump across `RIL_PARA` when I want text lines in text blocks.

That was fixed in https://github.com/OCR-D/ocrd_tesserocr/pull/158/commits/1ac011e1845afb08ec5a7fe74657b3deb7e65977 right?

> 1. keep image padding, but set defaults for both kinds of padding to 0px (as before)
> 2. remove image padding, start a new PR for it (and get involved in the GT measurements there)
> 3. expose a different parameter name for image padding, say image_padding, which defaults to 0px

What was the resolution here?

bertsky:
> > Thanks @stweil! I forgot I need to jump across `RIL_PARA` when I want text lines in text blocks.
> 
> That was fixed in [1ac011e](https://github.com/OCR-D/ocrd_tesserocr/commit/1ac011e1845afb08ec5a7fe74657b3deb7e65977) right?

Yes!

> > 1. keep image padding, but set defaults for both kinds of padding to 0px (as before)
> > 2. remove image padding, start a new PR for it (and get involved in the GT measurements there)
> > 3. expose a different parameter name for image padding, say image_padding, which defaults to 0px
> 
> What was the resolution here?

87e6444 implements option 1, which is okay with me, too.

We'll need some announcements and updates to the workflow guides when/after merging this, probably a dedicated call. Same for https://github.com/cisocrgroup/ocrd_cis/pull/77 (which is not quite ready yet). That's why I waited for you. We have to stir everything up a bit, but let's at least minimise the rumble by doing it all at once.

kba:
> We'll need some announcements and updates to the workflow guides when/after merging this, probably a dedicated call. Same for cisocrgroup/ocrd_cis#77 (which is not quite ready yet).

OK, let's discuss what needs to change documentation-wise in the call thursday and plan to have this, the ocrd_cis resegment PR and depending documentation changs ready for the open tech call next week? Is that realistic?

bertsky:
> OK, let's discuss what needs to change documentation-wise in the call thursday and plan to have this, the ocrd_cis resegment PR and depending documentation changs ready for the open tech call next week? Is that realistic?

Yes, agreed!

bertsky:
Hmm, sorry for adding yet another verse to the epic, but I must revisit the decision to give up the `overwrite_*` parameters (allowing only `True` from now on):

> Since the existing `overwrite_*` parameters did not really do anything useful (as they merely allowed redundant segments when `False`), we just ignore them from now on.

I forgot that there is in fact one valid use-case for these parameters being false, which is input which _sometimes_ has sub-segments, but sometimes does not (because the previous segmentation failed on that segment). For example, segment-line with `overwrite_lines=False` does make sense if I expect that regions sometimes contain lines already (where I don't want to add more), but sometimes do not (where I want to find some). That's not the same as the old behaviour (which would add segments regardless of existing ones), and not the same as the more desirable incremental behaviour (which would add segments only where no existing ones are located), but at least one reason for still allowing `False` even now.

Should I make the respective changes?

kba:
ok, if we do need `overwrite_*` to allow users to only "fill in the blanks" in inconsistently segmented data, then indeed please reintroduce them.

kba:
:tada:

PR #162, opened 2020-12-05
Polygonalize segments by shrinking to children

bertsky:
This implements an idea I had long time ago to at least get _some_ polygons out of Tesseract, reducing the large overlap between bounding boxes. If you set `shrink_polygons`, then `segment*` and `recognize` will post-process their segmentation by projecting convex hull upwards from `textequiv_level` to `segmentation_level`.

For example, `ocrd-tesserocr-segment` now yields:
tesserocr-polygons

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -92,7 +92,9 @@ class TesserocrRecognize(Processor):
         
         Set up Tesseract to recognise each segment's image (either from
         AlternativeImage or cropping the bounding box rectangle and masking
-        it from the polygon outline) with the appropriate mode and ``model``.
+        it from the polygon outline) with the appropriate segmentation mode
+        and ``model``. (If no ``model`` is given, only layout analysis will
+        be performed.)
         
         Next, if there still is a gap between the current level in the PAGE hierarchy
         and the requested ``textequiv_level``, then iterate down the result hierarchy,
@@ -100,7 +102,8 @@ class TesserocrRecognize(Processor):
         text line order, reading direction and orientation at the region/table level).
         
         Then, at ``textequiv_level``, remove any existing TextEquiv, unless
-        ``overwrite_text`` is false, and add text and confidence results .
+        ``overwrite_text`` is false, and add text and confidence results, unless
+        ``model`` is empty.
         
         The special value ``textequiv_level=none`` behaves like ``glyph``,
         except that no actual text recognition will be performed, only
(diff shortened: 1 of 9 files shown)


bertsky:
Note that the big downside of implementing `shrink_polygons` as a post-processor iterating through the PAGE hierarchy instead of right away through Tesseract hierarchy is that you _only_ get any actual polygons if you use multi-level segmentation (e.g. regions and lines and words). But in that case, you will _already_ avoid conflicting/overlapping lines. However, where shrinking would be most needed is when (for some reason) you can only employ Tesseract for a single level, e.g. `ocrd-tesserocr-segment-region` or `ocrd-tesserocr-segment-line`.

Here's an illustration:
- region segmentation only (nothing to shrink to; only bounding boxes)
tesserocr-polygons2-block
- region and line segmentation (shrinks regions to lines; coarse polygons)
tesserocr-polygons2-line
- region, line and word segmentation (shrinks lines to words, then regions to lines; better polygons)
tesserocr-polygons2-word
- region, line, word and glyph segmentation (shrinks words to glyphs, lines to words, then regions to lines; tight polygons)
tesserocr-polygons2

To get tight polygons when segmenting regions only, going top-down instead of bottom-up, we would have to query the Tesseract result iterator down to the glyph level for each block and get the hull of all segments, but then _reset_ the iterator to the start of the block. Unfortunately, Tesseract's API does not contain a reset function for the block level, only `RestartParagraph`, `RestartRow`, `BeginWord`. And since iterators are CPython objects, we also cannot just make a copy of them in another state...

bertsky:
> To get tight polygons when segmenting regions only, going top-down instead of bottom-up, we would have to query the Tesseract result iterator down to the glyph level for each block and get the hull of all segments, but then _reset_ the iterator to the start of the block. Unfortunately, Tesseract's API does not contain a reset function for the block level, only `RestartParagraph`, `RestartRow`, `BeginWord`. And since iterators are CPython objects, we also cannot just make a copy of them in another state...

Good news: I did find a solution for this. I emulated a `RestartBlock` (not defined in Tesseract) and `BeginWord` (not exposed in `tesserocr`) by simply resetting the iterator to the next higher level and moving forward to the current position again. Instead of post-processing PAGE projecting from level to level upwards, it now always takes the hull of the lowest PageIterator level (symbols) directly.

bertsky:
The bad news is that I also made a big conceptual mistake in #158: My parameterization `segmentation_level` for segmentation entry point and `textequiv_level` for segmentation exit point – as well as recognition oplevel – (with `none` for no recognition but maximal segmentation) makes it impossible to get single-level segmentation without recognition. So for example, `ocrd-tesserocr-segment-region` suddenly gives you lines, words and glyphs, but `ocrd-tesserocr-recognize -P segmentation_level region -P textequiv_level region` computes OCR and annotates text. Sorry about that!

Here's an idea how to fix this (without throwing it all away): We could separate the question of whether to do only segmentation (`AnalyseLayout()`) or segmentation plus recognition (`Recognize()`) from the `textequiv_level` parameter, and let the `model` parameter control this implicitly: If `model` is empty, then only segmentation, otherwise segmentation+recognition. (AFAIK the layout analysis can live with any model, so we could keep the current default, which loads the last installed model, or take `eng`.) So:
- `ocrd-tesserocr-segment-region` would use `textequiv_level='region' model=None`
- `ocrd-tesserocr-segment-line` would use `textequiv_level='line' model=None`
- `ocrd-tesserocr-segment-word` would use `textequiv_level='word' model=None`
- `ocrd-tesserocr-segment` would use `textequiv_level='word' model=None`
- `ocrd-tesserocr-recognize` would use `textequiv_level='word' model=last` by default or whatever you told it to

Opinions?

bertsky:
We should round this off before merging 0.10 in ocrd_all and adapting all documentation. @kba @stweil @wrznr

kba:
> The bad news is that I also made a big conceptual mistake in #158: My parameterization `segmentation_level` for segmentation entry point and `textequiv_level` for segmentation exit point – as well as recognition oplevel – (with `none` for no recognition but maximal segmentation) makes it impossible to get single-level segmentation without recognition. So for example, `ocrd-tesserocr-segment-region` suddenly gives you lines, words and glyphs, but `ocrd-tesserocr-recognize -P segmentation_level region -P textequiv_level region` computes OCR and annotates text. Sorry about that!

Oops. But good that you noticed before ocrd_all merge/documentation update.

> Here's an idea how to fix this (without throwing it all away): We could separate the question of whether to do only segmentation (`AnalyseLayout()`) or segmentation plus recognition (`Recognize()`) from the `textequiv_level` parameter, and let the `model` parameter control this implicitly: If `model` is empty, then only segmentation, otherwise segmentation+recognition. (AFAIK the layout analysis can live with any model, so we could keep the current default, which loads the last installed model, or take `eng`.) So:
> 
> * `ocrd-tesserocr-segment-region` would use `textequiv_level='region' model=None`
> * `ocrd-tesserocr-segment-line` would use `textequiv_level='line' model=None`
> * `ocrd-tesserocr-segment-word` would use `textequiv_level='word' model=None`
> * `ocrd-tesserocr-segment` would use `textequiv_level='word' model=None`
> * `ocrd-tesserocr-recognize` would use `textequiv_level='word' model=last` by default or whatever you told it to
> 
> Opinions?

Makes sense. If users want text recognition they will also explicitly provide the model(s) they want to use for recognition, so this is a reasonable convention. From my side, I'd be content with this segmentation_level/textequiv_level/model logic.

bertsky:
> Makes sense. If users want text recognition they will also explicitly provide the model(s) they want to use for recognition, so this is a reasonable convention. From my side, I'd be content with this segmentation_level/textequiv_level/model logic.

See d1ffd70. I still kept the `textequiv_level='none'` logic as an alternative.

ocrd_tesserocr/recognize.py
```diff
@@ -72,7 +73,7 @@ class TesserocrRecognize(Processor):
 
     def __init__(self, *args, **kwargs):
         kwargs['ocrd_tool'] = OCRD_TOOL['tools'][TOOL]
-        kwargs['version'] = OCRD_TOOL['version']
```
kba, line 75: Good idea! We should probably also do that for ocrd_calamari

kba:
I've tested this against a few examples from our GT and did not notice any problems with `shrink_polygons`, even noticed an error in the GT:

image

So from my point of view, this can be merged and we can update the documentation and ocrd_all.

PR #163, opened 2020-12-10
adjust output fileGrp name

cneud:
`OCR-D-OCR-FONTSTYLE` seems more specific than `OCR-D-OCR-STYLE` (to me at least). `TEXTSTYLE` or `FONTSHAPE` would also be fine. I understand this is probably just cosmetics, so treat this as a suggestion ;)


kba (approved):
It's only relevant for the `--help` output AFAICS, so the more expressive the name the better.

stweil (approved):

PR #166, opened 2020-12-22
Integration with OCR-D/spec#169 (resource manager)

kba:
This is just a proof-of-concept that it is possible to load tesseract models installed with `ocrd resmgr` into the cache directory.

The tricky part here is that there is only one `TESSDATA_PREFIX` but potentially multiple directories with models. So while it is no problem to look up models in various folders, only one of the can be used as the TESSDATA_PREFIX. Suggestions for a reasonable resolution to this dilemma are welcome.

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -63,6 +63,11 @@ TOOL = 'ocrd-tesserocr-recognize'
 CHOICE_THRESHOLD_NUM = 6 # maximum number of choices to query and annotate
 CHOICE_THRESHOLD_CONF = 0.2 # maximum score drop from best choice to query and annotate
 
+# XXX how to handle multiple search paths for models with a single # TESSDATA_PREFIX?
+from ocrd_utils import XDG_CACHE_HOME
+TESSDATA_PREFIX = os.path.join(XDG_CACHE_HOME, 'ocrd-tesserocr-recognize')
+
+
 def get_languages(*args, **kwargs):
     """
     Wraps tesserocr.get_languages() with a fixed path parameter.
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -203,17 +203,20 @@ class TesserocrRecognize(Processor):
         inlevel = self.parameter['segmentation_level']
         outlevel = self.parameter['textequiv_level']
         
-        model = get_languages()[1][-1] # last installed model
         if 'model' in self.parameter:
             model = self.parameter['model']
             for sub_model in model.split('+'):
+                if sub_model.endswith('.traineddata'):
+                    self.logger.warning("Model '%s' has a  .traineddata extension, removing. Please use model names without .traineddata extension" % sub_model)
+                    sub_model = sub_model.replace('.traineddata', '')
                 # XXX this will ensure that the models are downloaded if not already in MODEL_LOCATION
-                # TODO handle existing/missing .traineddata extension
-                self.resolve_resource(sub_model)
+                self.resolve_resource('%s.traineddata' % sub_model)
                 if sub_model not in get_languages()[1]:
                     raise Exception("configured model " + sub_model + " is not installed")
                 self.logger.info("Using model '%s' in %s for recognition at the %s level",
                                  model, get_languages()[0], outlevel)
+        else:
+            model = get_languages()[1][-1] # last installed model
         
         with PyTessBaseAPI(path=MODEL_LOCATION, lang=model) as tessapi:
             if outlevel == 'glyph':
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -63,11 +63,6 @@ TOOL = 'ocrd-tesserocr-recognize'
 CHOICE_THRESHOLD_NUM = 6 # maximum number of choices to query and annotate
 CHOICE_THRESHOLD_CONF = 0.2 # maximum score drop from best choice to query and annotate
 
-# XXX how to handle multiple search paths for models with a single # TESSDATA_PREFIX?
-from ocrd_utils import XDG_CACHE_HOME
-TESSDATA_PREFIX = os.path.join(XDG_CACHE_HOME, 'ocrd-tesserocr-recognize')
-
-
 def get_languages(*args, **kwargs):
     """
     Wraps tesserocr.get_languages() with a fixed path parameter.
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -56,7 +56,7 @@ from ocrd_models.ocrd_page_generateds import (
 from ocrd_modelfactory import page_from_file
 from ocrd import Processor
 
-from .config import TESSDATA_PREFIX, OCRD_TOOL
+from .config import MODEL_LOCATION, OCRD_TOOL
 
 TOOL = 'ocrd-tesserocr-recognize'
 
@@ -72,7 +72,7 @@ def get_languages(*args, **kwargs):
     """
     Wraps tesserocr.get_languages() with a fixed path parameter.
     """
-    return get_languages_(*args, path=TESSDATA_PREFIX, **kwargs)
+    return get_languages_(*args, path=MODEL_LOCATION, **kwargs)
 
 class TesserocrRecognize(Processor):
 
@@ -217,7 +217,7 @@ class TesserocrRecognize(Processor):
                 self.logger.info("Using model '%s' in %s for recognition at the %s level",
                                  model, get_languages()[0], outlevel)
         
-        with PyTessBaseAPI(path=TESSDATA_PREFIX, lang=model) as tessapi:
+        with PyTessBaseAPI(path=MODEL_LOCATION, lang=model) as tessapi:
             if outlevel == 'glyph':
                 # populate GetChoiceIterator() with LSTM models, too:
                 tessapi.SetVariable("lstm_choice_mode", "2") # aggregate symbols
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -209,8 +209,6 @@ class TesserocrRecognize(Processor):
                 if sub_model.endswith('.traineddata'):
                     self.logger.warning("Model '%s' has a  .traineddata extension, removing. Please use model names without .traineddata extension" % sub_model)
                     sub_model = sub_model.replace('.traineddata', '')
-                # XXX this will ensure that the models are downloaded if not already in TESSDATA_PREFIX
-                self.resolve_resource('%s.traineddata' % sub_model)
                 if sub_model not in get_languages()[1]:
                     raise Exception("configured model " + sub_model + " is not installed")
                 self.logger.info("Using model '%s' in %s for recognition at the %s level",
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -207,6 +207,9 @@ class TesserocrRecognize(Processor):
         if 'model' in self.parameter:
             model = self.parameter['model']
             for sub_model in model.split('+'):
+                # XXX this will ensure that the models are downloaded if not already in MODEL_LOCATION
+                # TODO handle existing/missing .traineddata extension
+                self.resolve_resource(sub_model)
                 if sub_model not in get_languages()[1]:
                     raise Exception("configured model " + sub_model + " is not installed")
                 self.logger.info("Using model '%s' in %s for recognition at the %s level",
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -56,7 +56,7 @@ from ocrd_models.ocrd_page_generateds import (
 from ocrd_modelfactory import page_from_file
 from ocrd import Processor
 
-from .config import MODEL_LOCATION, OCRD_TOOL
+from .config import TESSDATA_PREFIX, OCRD_TOOL
 
 TOOL = 'ocrd-tesserocr-recognize'
 
@@ -67,7 +67,7 @@ def get_languages(*args, **kwargs):
     """
     Wraps tesserocr.get_languages() with a fixed path parameter.
     """
-    return get_languages_(*args, path=MODEL_LOCATION, **kwargs)
+    return get_languages_(*args, path=TESSDATA_PREFIX, **kwargs)
 
 class TesserocrRecognize(Processor):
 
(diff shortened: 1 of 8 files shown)


bertsky (commented):
Thanks for setting the example and opening the discussion!

There are many sides to this. For one, tesserocr has no fixed `TESSDATA_PREFIX` itself, but allows a `path` kwarg to both `PyTessBaseAPI` and `get_languages`. So – iff the user requests a set of models all of which reside in the same directory – we could search our priority list of prefix paths, comparing the data files present in each one with the requested set, and pick the first full match. But what if the models are scattered across directories? What if the user does not know about all search paths and another directory with the wrong model version sneaks in?

At this point, IMHO we should strive to make the user experience / interface as intuitive and "correct" as possible, without much consideration how we are going to implement this in beauty. 

We might have to create temporary directories with symlinks / hardlinks / COW links / copies of all models in the search paths to achieve this. Or (in the opposite line of thought) prefer a single directory for all models just because it exists (even if empty).

So what is the right kind of behaviour? What are the use cases?

bertsky:
> So what is the right kind of behaviour? What are the use cases?

It's probably quite simple actually.

We have to touch the ocrd_all rules for Tesseract models anyway: the apt-get method is kinda broken and should be replaced with the existing download rules (independent of whether `tesseract` is active as a module). And the download rules would need to be adapted to our new `$VIRTUAL_ENV/share/$MODULE` convention (by replacing `TESSDATA := $(VIRTUAL_ENV)/share/tessdata` with `$(VIRTUAL_ENV)/share/ocrd_tesserocr`). 

And for users of native ocrd_tesserocr (i.e. without ocrd_all) we could provide rules based on `ocrd resmgr`. (In fact, we could rebase all ocrd_all rules on the centralized knowledge base in resmgr from now on, merely delegating to its download facilities, but that's another topic.)

Lastly, users who install Tesseract natively themselves will always know what to do. Likewise, users who want to run the native Tesseract CLI know how to use `--tessdata-dir`. 

So I think we can ignore the compiled/configured-in `TESSDATA_PREFIX` entirely (both for native CLI and for ocrd_tesserocr, both the PPA default `/usr/share/tesseract-ocr/4.00/tessdata` and the git source default `$VIRTUAL_ENV/share/tessdata`).

XDG is somewhat out of question here: It's okay for interactive users in a native installation, but in an encapsulated ocrd_all rule, maybe even in Docker? We cannot even be sure what the user / HOME will be then. Besides, naturally, an ocrd_all user might expect _some_ isolation of the installation directory from the rest of the system. So even if we use `ocrd resmgr` in ocrd_all, we should only use the VIRTUAL_ENV prefix for data paths.

It therefore comes down to how we can coalesce our model installation options across various module installation options, and then make ocrd_tesserocr default to that choice (while still offering the old `TESSDATA_PREFIX` environment override).

kba:
Thanks for the detailled analysis, very helpful.

I agree that we should replace the model download code in ocrd_all with resmgr.

> `TESSDATA_PREFIX = os.path.join(XDG_CACHE_HOME, 'ocrd-tesserocr-recognize')`

Agree, that is the most straightforward location, since it's the resmgr default. If users follow our (to-be-updated) documentation, this is where models will be and if users have more complex model storage needs, they are probably still better of symlinking the models to `$HOME/.cache`.

bertsky:
But how do we address the unreliability of `HOME` in the Docker context and the user expectation of data encapsulation under `VIRTUAL_ENV` in the native context? 

(I was actually more inclined towards `$(VIRTUAL_ENV)/share/ocrd_tesserocr`. We can always count on `VIRTUAL_ENV` being well defined – Python or non-Python module, Docker or native. Maybe this should even be the default for resmgr downloads ...)

kba:
`$(VIRTUAL_ENV)/share/ocrd_tesserocr` -> `$(VIRTUAL_ENV)/share/ocrd-tesserocr-recognize` but otherwise agreed.

bertsky:
So in the call we complemented the picture:
- Docker installation is problematic with the VIRTUAL_ENV/share default, because we cannot mount models there flat, but we definitely don't want to include the huge models into the Docker images, so some HOME (XDG) default or /usr/local would be better there
- users installing modules one by one natively without ocrd_all might not even use a common VIRTUAL_ENV, let alone define/activate it before installation, and they might be more inclined to expect models to stay under HOME (XDG)
- thus we need some mechanism of _configuring_ the default resource path at install-time; the best candidate is the `ocrd` package itself, which is always needed (Python and bashlib); so `core` install rules could create a config file as pkg_resource (like the `resource_list.yml` just introduced)

kba:
> because we cannot mount models there flat

One option to alleviate this problem would be to have an intermediary directory, say `ocrd-resources` where the actual per-processor directories reside:

`$XDG_CONFIG_HOME/ocrd-tesserocr-recognize` -> `$XDG_CONFIG_HOME/ocrd-resources/ocrd-tesserocr-recognize`

bertsky:
> > because we cannot mount models there flat
> 
> One option to alleviate this problem would be to have an intermediary directory, say `ocrd-resources` where the actual per-processor directories reside:
> 
> `$XDG_CONFIG_HOME/ocrd-tesserocr-recognize` -> `$XDG_CONFIG_HOME/ocrd-resources/ocrd-tesserocr-recognize`

Yes! Or `$VIRTUAL_ENV/share/ocrd-resources/ocrd-tesserocr-recognize`.

BTW, the only case where multiple processors need the same models is between ocrd-tesserocr-recognize (or its aliases) and ocrd-tesserocr-crop. The latter also needs eng.traineddata...

kba:
> Yes! Or $VIRTUAL_ENV/share/ocrd-resources/ocrd-tesserocr-recognize.

OK, I'll adapt the spec and implementation.

> BTW, the only case where multiple processors need the same models is between ocrd-tesserocr-recognize (or its aliases) and ocrd-tesserocr-crop. The latter also needs eng.traineddata...

I see but I think we can live with that, `eng.traineddata` is small enough to be duplicated and if users want to save every last byte, they can symlink it to the `ocrd-tesserocr-recognize` version of it.

kba:
> * thus we need some mechanism of _configuring_ the default resource path at install-time; the best candidate is the `ocrd` package itself, which is always needed (Python and bashlib); so `core` install rules could create a config file as pkg_resource (like the `resource_list.yml` just introduced)

I have not found a reliable and intuitive way to configure this behavior at runtime. I'll implement it as a config file `$HOME/.config/ocrd.yml` which stores the default resource storage location.

kba:
* The resource location in core is now configurable
* ocrd_tesserocr uses the default resource location instead of `TESSDATA_PREFIX`
* All tesserocr processors use the `ocrd-tesserocr-recognize` location, except `-crop`. Since we don't do any actual resolving in ocrd_tesserocr, it doesn't matter what the processor is called and we could reuse the same location for all processors - right?

bertsky (changes requested):
I am confused by this.

I thought we agreed to keep the `TESSDATA_PREFIX` environment variable as an override to the resmgr location?

Why an extra install path for crop now? We wanted to share models between processors here.

kba:
> Why an extra install path for crop now? We wanted to share models between processors here.

I was confused about that too, what I meant by

> Since we don't do any actual resolving in ocrd_tesserocr, it doesn't matter what the processor is called and we could reuse the same location for all processors - right?

is: I don't think we need an extra directory for `-crop` but it's late and I'm hungry and wanted to make sure :)

> I thought we agreed to keep the TESSDATA_PREFIX environment variable as an override to the resmgr location?

True, I'll adapt.

kba:
TESSDATA_PREFIX and single model location b2de1a7

ocrd_tesserocr/recognize.py
```diff
@@ -56,18 +56,23 @@
...
 TOOL = 'ocrd-tesserocr-recognize'
 
 CHOICE_THRESHOLD_NUM = 6 # maximum number of choices to query and annotate
 CHOICE_THRESHOLD_CONF = 0.2 # maximum score drop from best choice to query and annotate
 
+# XXX how to handle multiple search paths for models with a single # TESSDATA_PREFIX?
+from ocrd_utils import XDG_CACHE_HOME
+TESSDATA_PREFIX = os.path.join(XDG_CACHE_HOME, 'ocrd-tesserocr-recognize')
+
+
```
bertsky, line 70: Needs to be removed again?

bertsky (approved):
Much better, thanks – and let's call it a day!

ocrd_tesserocr/recognize.py
kba: Oops, right. 0bf2924

ocrd_tesserocr/recognize.py
```diff
@@ -207,12 +207,15 @@ def process(self):
         if 'model' in self.parameter:
             model = self.parameter['model']
             for sub_model in model.split('+'):
+                # XXX this will ensure that the models are downloaded if not already in MODEL_LOCATION
+                # TODO handle existing/missing .traineddata extension
+                self.resolve_resource(sub_model)
```
kba, line 212: This must be adapted once #167 is merged.

bertsky (commented):
BTW I don't see the merit of renaming the variable `TESSDATA_PREFIX` to `MODEL_LOCATION` everywhere. It's still the same mechanism as far as Tesseract is concerned, so it's the expected name here.

kba:
> BTW I don't see the merit of renaming the variable TESSDATA_PREFIX to MODEL_LOCATION everywhere. It's still the same mechanism as far as Tesseract is concerned, so it's the expected name here.

`TESSDATA_PREFIX` implies that it is the same mechanism that tesseract uses, with a with the compile-time default. But I don't mind renaming it back, upcoming. Currently, I'm fighting a bit with the fact that this directory can be empty of any models, that's why I'm force-pushing to get the CI right. I'll post a msg once that has been fixed.

bertsky:
> > BTW I don't see the merit of renaming the variable TESSDATA_PREFIX to MODEL_LOCATION everywhere. It's still the same mechanism as far as Tesseract is concerned, so it's the expected name here.
> 
> `TESSDATA_PREFIX` implies that it is the same mechanism that tesseract uses, with a with the compile-time default. But I don't mind renaming it back, upcoming.

Ok, you're right, it's a _different_ mechanism than the compile-time default, so perhaps we should indeed use a different name. Given the CLI's `tessdata-dir`, API's `datapath`, and tesserocr's `path` kwarg or `GetDatapath`, we can also call it something different altogether – I don't mind `MODEL_LOCATION` either.

kba:
CI is fixed. Doing `ocrd resmgr download ocrd-tesserocr-recognize deu.traineddata` before running the code did the trick.

> renaming the variable TESSDATA_PREFIX to MODEL_LOCATION everywhere

reverted in a223919

ocrd_tesserocr/recognize.py
bertsky: Yes. But I wonder how we can ensure this still only uses our single unique resource location (for which we even have an envvar override). `Processor.resolve_resource` would not only mix in other locations, it could even download to other locations. None of these other locations would actually work, however. So perhaps we should re-implement that function here. 

Or we drop it and just keep raising an exception if any model is not in `get_languages()`?

Also, perhaps we should move the model initialization code from `process` to some `setup` which we call from the constructor if a workspace was passed in. (So a future processor server or the current workflow server could pick it up.) See ocrd_cis for that pattern.

bertsky:
> > renaming the variable TESSDATA_PREFIX to MODEL_LOCATION everywhere
> 
> reverted in [a223919](https://github.com/OCR-D/ocrd_tesserocr/commit/a2239198c9b6878feaa031a50acc6d5447c45c8d)

Faster than my commenting :smile:

ocrd_tesserocr/recognize.py
kba: I think you're right, that we shouldn't be too clever for our own good here. Download-on-demand is nice and all but even more so is clarity. For the CI e.g. I resorted to just doing `ocrd resmgr download` explicitly instead of relying on this mechanism. So I'll just remove it, better to have a clear error message than a potentially buggy on-demand downloading.

kba: 44dd99d

kba: > Also, perhaps we should move the model initialization code from process to some setup which we call from the constructor if a workspace was passed in. (So a future processor server or the current workflow server could pick it up.) See ocrd_cis for that pattern.

Valid point but let's do this in a separate PR.

stweil:
Did I already mention that I don't like that environment variable `TESSDATA_PREFIX` and that I'd prefer to remove it from the Tesseract code? Which kind of handling would fit for OCR-D? Some features which are easy to implement:

1. support absolute and relative pathname for traineddata file in addition to short name for a language
2. support URL for traineddata file in addition to short name for a language

The first case could also be handled automatically with the URL case by using `file://` instead of `https://`.

ocrd_tesserocr/recognize.py
bertsky: > So I'll just remove it, better to have a clear error message than a potentially buggy on-demand downloading.

But then we need to make an exception in our documentation. So how about the other processors – do we really want automatic downloading of resources in a processor, now that your `ocrd resmgr` is super mighty?

(One can argue against that as a general convenience feature, by remarking that it is not guaranteed to work on all requested/configured resources anyway, and that workflow flaws need to be detected early-on independently. Especially if we want to separate setup and processing.)

bertsky:
> Did I already mention that I don't like that environment variable `TESSDATA_PREFIX` and that I'd prefer to remove it from the Tesseract code?

You mean the environment variable override here in ocrd_tesserocr, or the compile-time variable in Tesseract proper?

> Which kind of handling would fit for OCR-D?

The current tesserocr wrapper is sufficient IMHO – it does have a default, but a `path` kwarg everywhere.

The only thing Tesseract itself could do to improve the overall situation for OCR-D would be to allow loading models from multiple paths – so yes, allowing absolute and CWD-relative path names from Tesseract would be nice. (Tesserocr would probably not even have to be adapted.)

I don't think URL schemes are such a good thing in Tesseract, though.

kba:
> support absolute and relative pathname for traineddata file in addition to short name for a language

:+1: that would be helpful and would allow us to make full use of the resource management feature in ocrd_tesserocr same as the other processors.

ocrd_tesserocr/recognize.py
kba: The main argument against on-demand downloading IMHO is that it hides errors. I am right now debugging the resource-manager adaption in ocrd_anybaseocr and it doesn't find the models and tries to download them again, which is actually broken right now, as I just learned 😠 And you wouldn't get a nice progress bar or anything, it would just appear to hang while downloading gigabytes of redundant data.... So, I'll patch this feature out of the code and documentation, best to use `ocrd resmgr` exclusively for this.

kba:
@bertsky can I merge and release?

bertsky:
> can I merge and release?

@kba yes please

PR #167, opened 2021-01-21
log models with .traineddata extension as error

kba:
The problem is that `resmgr` doesn't know the `parameter_usage` of unregistered resources, since they are resource-specific, not processor-specific:

```
$ ocrd resmgr download -n ocrd-tesserocr-recognize https://github.com/tesseract-ocr/tessdata/raw/master/fin.traineddata
[...]
INFO ocrd.cli.resmgr - Use in parameters as 'fin.traineddata'
```

But that is the wrong parameter usage.

Remedies:

  1. Make `parameter_usage` processor-specific. That requires changing the data model which I am not a fan of.
  2. Warn users in tesserocr-recognize that they have a likely spurious `.traineddata` extension (what this PR currently does)
  3. Chop of the `.traineddata` form any (sub-)models in tesserocr.

I don't think there should ever be a model named `foo.traineddata.traineddata`, so I think 3 would be a valid workaround.

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -207,6 +207,9 @@ class TesserocrRecognize(Processor):
         if 'model' in self.parameter:
             model = self.parameter['model']
             for sub_model in model.split('+'):
+                if sub_model.endswith('.traineddata'):
+                    self.logger.warning("Model '%s' has a  .traineddata extension, removing. Please use model names without .traineddata extension" % sub_model)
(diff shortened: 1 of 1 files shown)


bertsky:
Yes. I'd say combine 2 and 3, i.e. warn that we chopped off the extension because we assumed it must be incorrect.

kba:
> Yes. I'd say combine 2 and 3, i.e. warn that we chopped off the extension because we assumed it must be incorrect.

done 79e51a8

PR #170, opened 2021-02-12
Skip recognition when text exists and not `overwrite_text`

bertsky:
This redefines the meaning of `overwrite_text`: If text exists for a segment, then skip it (instead of always running the recognition and just appending new results).

### DIFF
--- a/ocrd_tesserocr/ocrd-tool.json
+++ b/ocrd_tesserocr/ocrd-tool.json
@@ -172,6 +172,12 @@
           "default": {},
           "description": "Dictionary of additional Tesseract runtime variables (cf. tesseract --print-parameters), string values."
         },
+        "oem": {
+          "type": "string",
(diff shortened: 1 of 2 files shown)


ocrd_tesserocr/ocrd-tool.json
```diff
@@ -172,6 +172,12 @@
           "default": {},
           "description": "Dictionary of additional Tesseract runtime variables (cf. tesseract --print-parameters), string values."
         },
+        "oem": {
```
kba, line 175: We could add that to the version string added as metadata to the PAGE-XML, so it would be something like `0.11.0 (tesseract 5.0.0-alpha-20201231 - TESSERACT_LSTM_COMBINED)`.

kba (approved):

ocrd_tesserocr/ocrd-tool.json
bertsky: yes, but it's already in the `parameter` metadata.

kba: True.

Issue #69, opened 2019-08-05
recognize: use primaryScript or TextStyle to load model


bertsky:
In the current state, the OCR model has to be selected in the fixed parameter JSON for the whole pipeline (all pages, all regions, all lines). We should at least offer a setting like `dynamic` that instead looks into ...
- `mods:language` of the workspace's METS file
- `@primaryScript` and `@secondaryScript` of the elements to be processed (or their parents), depending on `textequiv_level`
- `TextStyle/@fontFamily` of the elements to be processed (or their parents), depending on `textequiv_level` – as described by the [spec](https://ocr-d.github.io/page#textstyle)

...and combines this information _somehow_ to select one of the predefined models. (Predefined could include custom built models, though. So maybe this must be more than a single new value in the parameter file.)

bertsky:
> ...and combines this information _somehow_ to select one of the predefined models. (Predefined could include custom built models, though. So maybe this must be more than a single new value in the parameter file.)

One idea to get this configured would be a (partial) mapping between ISO 15924 (script) / ISO 639 (language) / [OCR-D](https://ocr-d.de/en/spec/page#clusters-of-typesets) (font) to Tesseract models (in the usual notation) as parameter. A mapping for the empty string could become the manual fall-back. Matches in multiple categories (language/script/font), or across multiple levels (page/region/line), could be mixed via `+` in the result. 

(Mappings of `type: object` are allowed syntactically in OCR-D's parameter JSON now.)

So for example, I could first run `ocrd-typegroups-classifier` (for font detection) and/or `ocrd-tesserocr-deskew` (for script detection), and then call `ocrd-tesserocr-recognize -P model_map '{ "German": "deu+Latin", "deu": "deu+Latin", "Latin": "lat+Latin", "Latn": "Latin", "Latf": "GT4HistOCR+ONB+Fraktur+frk", "Greek": "grc+ell+Greek", "Grek": "Greek", "Hebr": "Hebrew", "": "eng" }'`. 

Probably we also should introduce some `model_conf` threshold here.

Unfortunately, due to Tesseract's API, the implementation would need to re-initialize Tesseract each time a segment has a different script/language/font annotation than the previous. But one could control this performance/quality trade-off by running detection on regions or pages only.

(A problem that first needs to be addressed though is the [formalization](https://github.com/PRImA-Research-Lab/PAGE-XML/issues/27) of script and language identifications in PAGE.)

bertsky:
[...]

PR #175, opened 2021-03-16
OSD on line level, recognition by loading script or lang from PAGE

bertsky:
In partial fulfillment of #69 – but I'm afraid osd.traineddata is just too bad for script detection to make this work reliably. Latin vs Cyrillic vs Arabic vs Chinese etc might be easy, but Greek for example does not work on the line level... 

Nevertheless, once we do have script / font detectors on the line level, we could make this work.

- [x] update the PR to the current master
- [x] ~~implement `style_model` (map from `TextStyle` alists to model name)~~
- [x] implement `xpath_model` / `xpath_parameters` (map from PAGE-XML XPath queries to model / variable settings)
- [x] implement `auto_model` (try each loaded model and pick best scoring one)

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -60,8 +60,8 @@ from .config import TESSDATA_PREFIX, OCRD_TOOL
 
 TOOL = 'ocrd-tesserocr-recognize'
 
-CHOICE_THRESHOLD_NUM = 6 # maximum number of choices to query and annotate
-CHOICE_THRESHOLD_CONF = 0.2 # maximum score drop from best choice to query and annotate
+CHOICE_THRESHOLD_NUM = 16 # maximum number of choices to query and annotate
+CHOICE_THRESHOLD_CONF = 0.001 # maximum score drop from best choice to query and annotate
 
 def get_languages(*args, **kwargs):
     """
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -343,8 +343,6 @@ class TesserocrRecognize(Processor):
                         # disable table detection here, so tables will be
                         # analysed as independent text/line blocks:
                         tessapi.SetVariable("textord_tabfind_find_tables", "0")
-                    if not segment_only:
-                        self._reinit(tessapi, page)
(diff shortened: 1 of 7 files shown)


ocrd_tesserocr/recognize.py
```diff
@@ -1190,6 +1196,30 @@ def page_update_higher_textequiv_levels(level, pcgts):
                     region_conf /= len(lines)
             region.set_TextEquiv( # replace old, if any
                 [TextEquivType(Unicode=region_unicode, conf=region_conf)])
+    def reinit(self, tessapi, segment):
+        tag = segment.__class__.__name__[:-4]
+        ident = segment.id if hasattr(segment, 'id') else segment.imageFilename
```
kba, line 1201: Can be just `segment.id` with https://github.com/OCR-D/core/pull/683

kba (approved):
Very good, thank you! The mapping can only be as good as `osd` is in detecting script and language, of course, but it's a step towards dynamic model selection and we can use the patterns here in other processors.

bertsky:
> The mapping can only be as good as `osd` is in detecting script and language, of course, 

Now goes beyond that: you can use e.g. `-P script_model '{ "Latn - Latin": "lat+Latin", "Grek - Greek": "grc+Greek" }' -P auto_model true -P model lat+Latin+grc+Greek` to have it try:
1. if `@primaryScript="Latn - Latin"`, then activate `lat+Latin`
2. elif `@primaryScript="Grek - Greek"`, then activate `grc+Greek`
3. otherwise try each loaded model individually and __pick the best-scoring one__ per segment

> but it's a step towards dynamic model selection and we can use the patterns here in other processors.

Yes, but that kind of dynamics will look quite different for each processor implementation (and no other engine has the model variety of Tesseract). I wish we could express any of that on the workflow level...

bertsky:
Note: depends on OCR-D/core#699

bertsky:
@kba, I still had to update our deployment rules for the resmgr changes. CI is now only failing because of the dependency on https://github.com/OCR-D/core/pull/699 – should I wait for your merge (and rebuild of `ocrd/core` on DH), or rather update to core's etree branch here?

Issue #174, opened 2021-03-16
list-resources not correct


bertsky:
After we resmgrized ocrd_tesserocr in #166, running any of the CLIs with `-L|--list-resources` is supposed to show the exact list of models available. However, since we cannot and did not adopt the scheme with multiple resource locations, but instead use only a single directory (OcrdResourceManager's default, which is `XDG_DATA_HOME/ocrd-resources/EXECUTABLE`) and allow overriding it via shell variable for compatibility reasons (`TESSDATA_PREFIX`), the default implementation in `ocrd_utils.list_all_resources` does not work here.

Thus, we should extend the constructor of all ocrd_tesserocr's processors to deal with `list_resources=True` in its own way (by using `.config.get_tessdata_path()`).

kba:
How do you mean that it does not work? It does list the resources in `XDG_DATA_HOME/ocrd-resources/ocrd-tesserocr-recognize` for `ocrd-tesserocr-recognize -L`. You mean that it does not correctly handle `TESSDATA_PREFIX`?

bertsky:
Yes, it does _not_ respect `TESSDATA_PREFIX` and still shows files under `/usr/local/share/ocrd-resources` and CWD, which will in fact _not_ be available. Delegating to `.config.get_tessdata_path` would fix that. (Probably applies to `--show-resource`, too.)

kba:
With https://github.com/OCR-D/spec/pull/181 merged and implemented in core, the restriction on location can be expressed as

```yaml
tools:
  ocrd-tesserocr-recognize:
    resource_locations: ['data']
```

`list_all_resources` can then be extended to take a list of locations to look in from the ocrd-tool.json and only list those.

We'll still need custom code in here to handle `TESSDATA_PREFIX` though so I am not sure whether it's worth it since ocrd_tesserocr is the only processor which would have a differing `resource_locations` :/

PR #176, opened 2021-03-18
custom --list-resource handler in init, #174

kba:
With this `--list-resources` will respect `TESSDATA_PREFIX` by listing all files/folders in `get_tessdata_path()`.

BTW we should consider a package `ocrd_testing` so we don't need to copy `tests/base.py` around.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
 
   build-python36:
     docker:
-      - image: ocrd/core
+      - image: python:3.6
     steps:
       - run: apt-get update && apt-get install -y --no-install-recommends make git curl
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 executors:
   ubuntu1804:
     docker:
-      - image: ubuntu:18.04
+      - image: ocrd/core
 
 commands:
   install-test-coverage:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ commands:
         type: string
     steps:
       - run: apt-get update
-      - run: apt-get install -y software-properties-common
+      - run: apt-get install -y software-properties-common imagemagick
       - run: add-apt-repository -y ppa:deadsnakes/ppa
       - run: apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Europe/Berlin apt-get install -y --no-install-recommends make git curl python<<parameters.python-version>>
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,17 +6,69 @@ jobs:
 
   build-python36:
     docker:
-      - image: python:3.6
+      - image: ubuntu:18.04
     steps:
-      - run: apt-get update && apt-get install -y --no-install-recommends make git curl
+      - run: apt-get update
+      - run: apt-get install -y software-properties-common
+      - run: add-apt-repository -y ppa:deadsnakes/ppa
+      - run: apt-get update && apt-get install -y --no-install-recommends make git curl python3.6
       - checkout
       - run: make deps-ubuntu
-      - run: make install
-      - run: make test-cli
-      - run: make coverage
+      - run: make install PYTHON=python3.6
+      - run: make test-cli PYTHON=python3.6
+      - run: make coverage PYTHON=python3.6
+      - codecov/upload
+
+  build-python37:
+    docker:
+      - image: ubuntu:18.04
+    steps:
+      - run: apt-get update
+      - run: apt-get install -y software-properties-common
+      - run: add-apt-repository -y ppa:deadsnakes/ppa
+      - run: apt-get update && apt-get install -y --no-install-recommends make git curl python3.7
+      - checkout
+      - run: make deps-ubuntu
+      - run: make install PYTHON=python3.7
+      - run: make test-cli PYTHON=python3.7
+      - run: make coverage PYTHON=python3.7
+      - codecov/upload
+
+
+  build-python38:
+    docker:
+      - image: ubuntu:18.04
+    steps:
+      - run: apt-get update
+      - run: apt-get install -y software-properties-common
+      - run: add-apt-repository -y ppa:deadsnakes/ppa
+      - run: apt-get update && apt-get install -y --no-install-recommends make git curl python3.8
+      - checkout
+      - run: make deps-ubuntu
+      - run: make install PYTHON=python3.8
+      - run: make test-cli PYTHON=python3.8
+      - run: make coverage PYTHON=python3.8
+      - codecov/upload
+
+  build-python39:
+    docker:
+      - image: ubuntu:18.04
+    steps:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -6,17 +6,69 @@ jobs:
 
   build-python36:
     docker:
-      - image: ocrd/core
+      - image: ubuntu:18.04
```
bertsky, line 9: Why not `ocrd/core` as base image anymore?

Now we have to repeat many steps from core's Dockerfile here, and I don't like the idea of having to rebase `FROM ubuntu:18.04` in each and every module in the future...

.circleci/config.yml
```diff
@@ -6,17 +6,69 @@ jobs:
...
     steps:
-      - run: apt-get update && apt-get install -y --no-install-recommends make git curl
+      - run: apt-get update
+      - run: apt-get install -y software-properties-common
+      - run: add-apt-repository -y ppa:deadsnakes/ppa
```
bertsky, line 13: Why `ppa:deadsnakes/ppa`, Python 3.9 perhaps?

.circleci/config.yml
```diff
@@ -6,17 +6,69 @@ jobs:
...
       - run: make deps-ubuntu
-      - run: make install
-      - run: make test-cli
-      - run: make coverage
+      - run: make install PYTHON=python3.6
+      - run: make test-cli PYTHON=python3.6
+      - run: make coverage PYTHON=python3.6
+      - codecov/upload
+
+  build-python37:
```
bertsky, line 22: Could this be made more elegant in Circle CI by just setting the `PYTHON` variable differently without repeating the same step definitions each time?

.circleci/config.yml
```diff
@@ -6,17 +6,69 @@ jobs:
...
+      - codecov/upload
+
+  build-python39:
+    docker:
+      - image: ubuntu:18.04
+    steps:
+      - run: apt-get update
+      - run: apt-get install -y software-properties-common
+      - run: add-apt-repository -y ppa:deadsnakes/ppa
+      - run: apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Europe/Berlin apt-get install -y --no-install-recommends make git curl python3.9
```
bertsky, line 60: Why `TZ=Europe/Berlin`, why only here?

ocrd_tesserocr/recognize.py
```diff
@@ -74,6 +77,12 @@ class TesserocrRecognize(Processor):
     def __init__(self, *args, **kwargs):
         kwargs['ocrd_tool'] = OCRD_TOOL['tools'][TOOL]
         kwargs['version'] = OCRD_TOOL['version'] + ' (' + tesseract_version().split('\n')[0] + ')'
+        if kwargs.get('list_resources', False):
```
bertsky, line 80: What about `show_resource`? We cannot use the superclass's constructor for that either, because `list_resource_candidates` does not apply...

bertsky (commented):
Many thanks for taking care of this!

.circleci/config.yml
kba: Main reason was the logging configuration in the Docker image which interferes with the testing (there was a file `ocrd.log` that was falsely listed as a resource of ocrd-tesserocr-recognize). We should debug this and change it back to `ocrd/core`, I was mostly trying to get the multi-version testing working with lots of trial and error.

.circleci/config.yml
kba: Yes, this offers a clean selection of all minor python versions, to make it easy to test all versions.

.circleci/config.yml
kba: CircleCI probably has some mechanism for build matrices, I'll look into it. While I agree that copy/pasting isn't very elegant, we (hopefully) don't have to change the configuration all that often so the redundancy isn't that big of a problem IMHO.

.circleci/config.yml
kba: Because https://docs.python.org/3/library/zoneinfo.html was added which makes the `python3.9` package install tzinfo which causes a prompt to choose a timezone.

ocrd_tesserocr/recognize.py
kba: Ah, right, I did test it but not with the `TESSDATA_PREFIX` said. I'll add that.

.circleci/config.yml
kba: https://github.com/OCR-D/ocrd_tesserocr/pull/176/commits/a14d73c8b840356ea01026268c9ff1385a7b17e9

.circleci/config.yml
kba: Is now set for all versions.

ocrd_tesserocr/recognize.py
kba: added in 59bb925. If you have an idea how the commented-out test might be coaxed into actually capturing STDOUT, pls let me know.

bertsky: > If you have an idea how the commented-out test might be coaxed into actually capturing STDOUT, pls let me know.

I guess `TestCase.invoke_cli` expects line-buffered output, whereas `shutil.copyfileobj` gives us unbuffered output.

.circleci/config.yml
bertsky: It should be possible to switch back to `ocrd/core` now, right?

bertsky: It's also urgent, because with the recent changes in core, we cannot just do `pip install ocrd` anymore:
```
FileNotFoundError: [Errno 2] No such file or directory: 'identify': 'identify'
```

ocrd_tesserocr/recognize.py
bertsky: But TBH I don't understand how these Pytest capturing fixtures (your use of `capfd` in `invoke_cli`) work.

bertsky:
CI currently fails for `ocrd.log` showing up as a resource candidate. This should probably be gone with an update to core or whereever we restrict resources to `.traineddata`.

bertsky:
@kba with https://github.com/OCR-D/core/pull/800 merged, what's the status here?

.circleci/config.yml
kba: resolved in https://github.com/OCR-D/ocrd_tesserocr/pull/176/commits/f01cd274b0bd965f8d12cf35111d6d080fc09014

ocrd_tesserocr/recognize.py
kba: I tried various possible fixes but to no avail. It doesn't make sense because we use the same pattern e.g. in `tests/cli/test_resmgr.py` in OCR-D/core and there are no issues there, they're using the same click and pytest.

But I verified manually that the specified resource is properly written to STDOUT for `-C` and the list is correct for `-L`.

Perhaps @M3ssman, being the pytest expert, has an idea?

But it should not stop us from releasing.

kba:
> @kba with OCR-D/core#800 merged, what's the status here?

Should have had a release last week, will have a release this weekend.

PR #177, opened 2021-06-26
Clipped page image

bertsky:
During layout analysis on the page level, Tesseract internally detects images and separators. Their bounding boxes can be queried via iterators, but the precise polygons are not available, and for sparse mode, only text blocks are returned. Moreover, we usually do not suppress separators and images in the foreground of consumers, even if we do annotate them in PAGE. This often yields suboptimal results, especially for OCRs like Ocropy or Calamari.

This PR therefore queries the internal binarized and nontext-suppressed (i.e. clipped) page-level image and annotates it as page-level derived image (with additional features `binarized,clipped`). As long as consumers do not opt out of the usual AlternativeImage retrieval priority, they will therefore get images without nontextual foreground.

### DIFF
--- a/ocrd_tesserocr/binarize.py
+++ b/ocrd_tesserocr/binarize.py
@@ -65,30 +65,47 @@ class TesserocrBinarize(Processor):
                 page_image, page_xywh, _ = self.workspace.image_from_page(
                     page, page_id)
                 LOG.info("Binarizing on '%s' level in page '%s'", oplevel, page_id)
-                
-                regions = page.get_TextRegion() + page.get_TableRegion()
-                if not regions:
-                    LOG.warning("Page '%s' contains no text regions", page_id)
-                for region in regions:
-                    region_image, region_xywh = self.workspace.image_from_segment(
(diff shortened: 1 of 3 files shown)


kba (approved):
IIUC this will prevent detection of false positives (regions w/o text) downstream, so LGTM.

bertsky:
> IIUC this will prevent detection of false positives (regions w/o text) downstream, so LGTM.

No, it's not about non-text regions. It is about non-textual (image/separator) foreground parts in text segments across the hierarchy, from text regions down to glyphs (depending on where the images are consumed and whether the consumers use the clipped page-level AlternativeImage annotated here). And it only applies when using ocrd-tesserocr-segment*/recognize with `segmentation_level=region` (i.e. as top level segmenter).

For example, if you run tesserocr-segment followed by calamari-recognize, this PR ensures that Calamari gets to see textline images without intruding h/v-lines (at least as far as Tesseract detects them correctly). Or if you use tesserocr-segment-region followed by ocropy-segment, the line segmenter should see clean text blocks.

PR #179, opened 2021-07-05
Improve tessapi reset

bertsky:
This makes it possible to pass in parameters like `user_words_file` or `user_patterns_file` (even in the `xpath_parameters` way), which are only evaluated during model initialization. 

It also removes the performance penalty cast by blanket per-segment `reset` in #175.

Plus it fixes a glitch in 318cefd that made all choices disappear when running `textequiv_level=glyph`.

### DIFF
--- a/ocrd_tesserocr/recognize.py
+++ b/ocrd_tesserocr/recognize.py
@@ -61,8 +61,9 @@ from .config import get_tessdata_path, OCRD_TOOL
 
 TOOL = 'ocrd-tesserocr-recognize'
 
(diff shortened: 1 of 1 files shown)


kba (approved):
LGTM. We should add tests for the (re-)initalization and `xpath_parameters` logic at some point.

bertsky:
> We should add tests for the (re-)initalization and `xpath_parameters` logic at some point.

Definitely. And then refactor to make the code readable again. I just don't have the time...

PR #184, opened 2022-02-14
ocrd-tool: list known resources

kba:


ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
           "default": "DEFAULT",
           "description": "Tesseract OCR engine mode to use:\n* Run Tesseract only - fastest,\n* Run just the LSTM line recognizer. (>=v4.00),\n*Run the LSTM recognizer, but allow fallback to Tesseract when things get difficult. (>=v4.00),\n*Run both and combine results - best accuracy."
         }
-      }
+      },
+      "resource_locations": ["data", "system", "module"],
```
bertsky, line 206:
```suggestion
      "resource_locations": ["module"],
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "name": "ONB.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract LSTM model based on Austrian National Library newspaper data",
+          "size": 4358948
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/equ.traineddata",
+          "name": "equ.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract equ model",
```
bertsky, line 226:
```suggestion
          "description": "Tesseract legacy model for mathematical equations",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 2251950
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/osd.traineddata",
+          "name": "osd.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract osd model",
```
bertsky, line 233:
```suggestion
          "description": "Tesseract legacy model for orientation and script detection",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 10562727
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/eng.traineddata",
+          "name": "eng.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract eng model",
```
bertsky, line 240:
```suggestion
          "description": "Tesseract LSTM model for contemporary (computer typesetting and offset printing) English",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 4113088
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/deu.traineddata",
+          "name": "deu.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract deu model",
```
bertsky, line 247:
```suggestion
          "description": "Tesseract LSTM model for contemporary (computer typesetting and offset printing) German",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 1525436
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/frk.traineddata",
+          "name": "frk.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract frk model",
```
bertsky, line 254:
```suggestion
          "description": "Tesseract LSTM model for historical (Fraktur typesetting and letterpress printing) German",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 6423052
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/script/Fraktur.traineddata",
+          "name": "Fraktur.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract Fraktur model",
```
bertsky, line 261:
```suggestion
          "description": "Tesseract LSTM model for historical Latin script with Fraktur typesetting",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 10915632
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tessdata_fast/raw/main/script/Latin.traineddata",
+          "name": "Latin.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract Latin model",
```
bertsky, line 268:
```suggestion
          "description": "Tesseract LSTM model for contemporary and historical Latin script",
```

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 89384811
+        },
+        {
+          "url": "https://github.com/tesseract-ocr/tesseract/archive/main.tar.gz",
```
bertsky, line 272: Interesting – I did not know this exists.

But shouldn't we be able to expect any realistic installation of Tesseract to already contain this preinstalled in the module directory (i.e. under TESSDATA_PREFIX)?

bertsky (commented):
Could we rebase that branch to direct it against #176, so that we can test everything new together?

ocrd_tesserocr/ocrd-tool.json
kba: Probably. I just wanted to make sure that the configs are also installable via `resmgr`. If you don't think it's needed, we can drop it.

bertsky: Let's keep it for now (but in the future we should try to find a way to identify that archive's content with the config files that usually get installed by Tesseract).

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -202,7 +202,81 @@
...
+          "size": 1058487
+        },
+        {
+          "url": "https://ub-backup.bib.uni-mannheim.de/~stweil/ocrd-train/data/ONB/tessdata_best/ONB_1.195_300718_989100.traineddata",
+          "name": "ONB.traineddata",
+          "parameter_usage": "without-extension",
+          "description": "Tesseract LSTM model based on Austrian National Library newspaper data",
+          "size": 4358948
+        },
+        {
```
bertsky, line 222: more up-to-date model:
```suggestion
        {
        "url": "https://ub-backup.bib.uni-mannheim.de/~stweil/tesstrain/frak2021/tessdata_best/frak2021-0.905.traineddata",
          "name": "frak2021.traineddata",
          "parameter_usage": "without-extension",
          "description": "Tesseract LSTM model based on Austrian National Library newspaper data",
          "size": 3421140
        },
        {
```

bertsky (commented):
@kba, with https://github.com/OCR-D/ocrd_tesserocr/pull/184 already merged, we should fast-track this before anyone runs into trouble.

LGTM – except that we now need to depend on `ocrd>=2.37`.

ocrd_tesserocr/ocrd-tool.json
```diff
@@ -207,7 +207,7 @@
           "description": "Tesseract OCR engine mode to use:\n* Run Tesseract only - fastest,\n* Run just the LSTM line recognizer. (>=v4.00),\n*Run the LSTM recognizer, but allow fallback to Tesseract when things get difficult. (>=v4.00),\n*Run both and combine results - best accuracy."
         }
       },
-      "resource_locations": ["module"],
+      "resource_locations": ["data"],
```
bertsky, line 210: @kba why is that?

kba: Because otherwise, the installed models in `$HOME/.local/share/ocrd-resources/ocrd-tesserocr-recognize` were not picked up. What would make sense in general would be `['data', 'module']` so that both the resources in the `data` location as well as module-bundled resources would be found. But that does not make sense for tesserocr either because we need a single place for all the resources.

bertsky: But IIRC the whole point for us was that in Tesseract's case, the resource location is module-defined, because we cannot lookup multiple locations, we must support the TESSDATA_PREFIX envvar, and we also would like to keep supporting the default datadir for Tesseract standalone at build-time.

If `module` is set, then it's actually correct _not_ to look into the `data` path (unless that coincides with the `module` path) IMO. On the contrary: It was an error of `resmgr download` to put the files there in the first place!

In … https://github.com/OCR-D/ocrd_tesserocr/blob/0427f6f31070bdc782a27e71cd6ac88142da9924/ocrd_tesserocr/config.py#L10-L13

… we delegate the `module` directory to `OcrdResourceManager`'s default, which would be the `data` path again. (We could however decide to use the install-time path from Tesseract compile-time here.)

So what we probably missed was to ask the module just where its module path is, so we can **write** new models there. We could also then rewrite the Processor constructor for `--list-resources` and `--show-resource` to **read** models there (so ocrd_tesserocr would not need a custom constructor anymore).

AFAICS all we need is an additional value `module` under `ocrd_utils.RESOURCE_LOCATIONS`, and resolving that value via `Processor.moduledir` and delegating to that in `OcrdResourceManager.location_to_resource_dir`.

Moreover, we should probably change `OcrdResourceManager.default_resource_dir` to query the concrete processor's `resource_locations` and use the first entry of that list instead of just `data`, and also to delegate to that function in `ocrd.cli.resmgr.download` for the `--location` option instead of just `data`.

kba: > So what we probably missed was to ask the module just where its module path is,

Yes! That seems to be the missing piece which confused me. Thanks for digging into this, I will open a PR aiming for proper support of `module` location.

PR #187, opened 2022-08-18
Drop custom tessdata location, properly support moduledir

kba:
This is the companion to OCR-D/core#904, doing away with overriding the path to `tessdata` and instead using the `self.moduledir` mechanism to tell core where it expects models to be stored.

~~Draft because OCR-D/core#904 needs to be finished before properly testing this.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,6 @@ deps:
 
 # Install Python deps for test via pip
 deps-test:
-	$(PIP) install -U pip
 	$(PIP) install -r requirements_test.txt
 
 # Build docker image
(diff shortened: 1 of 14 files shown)


bertsky:
So the situation with CI failure is this: despite the recent changes to test_cli, we still cannot properly simulate overriding the `TESSDATA_PREFIX` envvar. It does work on the CLI, but within pytest, it seems the in-line module initialization somehow gets called before the monkeypatching.

(And it does not help running with `--import-mode=append` or `--import-mode=importlib`, either.)

So what do we do? Skip this kind of test? Use some other mechanism for running the CLI?

kba:
> So what do we do? Skip this kind of test? Use some other mechanism for running the CLI?

We could set the envvar in the `make test` call, like we do for isolated logging tests in core - not the most elegant solution but probably easier to debug than the module-loading/pytest/monkeypatch mechanisms.

bertsky:
Note: I found out that it's the other test modules which drag in `tesserocr` initialization before any test gets run. (Just copy `test_cli.py` somewhere else and let it run alone.) Sadly, knowing this, I still have found no way to de-initialize this module (probably because it's Cython). For example,
```python
    monkeypatch.delitem(sys.modules, 'tesserocr')
    monkeypatch.delitem(sys.modules, 'ocrd_tesserocr')
```
does not help. Neither does using `importlib.reload`.

bertsky:
> We could set the envvar in the `make test` call, like we do for isolated logging tests in core - not the most elegant solution but probably easier to debug than the module-loading/pytest/monkeypatch mechanisms.

That would not allow setting isolated temporary directories, though. (Also, by setting `TESSDATA_PREFIX` for _all_ tests, the other tests would not find our true model files.)

In hope of fixing this properly some day, I just used the following workaround: running the tmpdir-based test_cli separate from the other tests.

PR #197, opened 2024-01-31
docker/install: build Tesseract from source

joschrew:
This PR is part of series to offer single ocrd modules as Docker Containers (ocrd slim containers) to be used with ocr-d network.

~~This Dockerfile currently doesn't work in all cases and it still needs updates. I created the PR anyway because I use/need it for my tests.~~ **EDIT** now works. (This basically migrates all the `install-tesseract` rules from ocrd_all's makefile here, where it actually belongs.)

My idea was to maybe create the tesseract Container with ocrd_all:
```
cd ocrd_all
git submodule update --init tesserocr/ core/ tesseract/ ocrd_tesserocr/
docker build --build-arg="OCRD_MODULES=core ocrd_tesserocr tesseract tesserocr " --no-cache -t my-ocrd-slim-container .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,65 @@
-FROM ocrd/core
+FROM ocrd/core:v2.62.0 AS base
+# set proper locales
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
+
+# set TESSDATA_PREFIX
+ENV TESSDATA_PREFIX /usr/local/share/tessdata
+
+ARG TESSERACT_REPOSITORY=https://github.com/tesseract-ocr/tesseract.git
+ARG TESSERACT_REF=5.3.3
+
+# set frontend non-interactive to silence interactive tzdata config
+ARG DEBIAN_FRONTEND=noninteractive
+
+
+# install common tools and tesseract build dependencies
+# use provided leptonica
+# tzdata required for proper timezone settings
+RUN apt-get update && apt-get install -y \
+	apt-utils \
+	build-essential \
+	g++ \
+	git \
+	libjpeg-dev \
+	libgif-dev \
+	libwebp-dev \
+	libopenjp2-7-dev \
+	libpng-dev \
+	libtiff-dev \
+	libtool \
+	pkg-config \
+	tzdata \
+	xzgv \
+	zlib1g-dev \
+	libleptonica-dev \
+	libpango1.0-dev \
+	libicu-dev \
+	autotools-dev \
+	automake \
+	libcurl4-nss-dev \
+	libarchive-dev
+
+# set proper date and timezone in container
+RUN echo "Europe/Berlin" > /etc/timezone
+RUN ln -sf /usr/share/zoneinfo/Europe/Berlin /etc/localtime
+RUN dpkg-reconfigure -f noninteractive tzdata
+
+# diagnostic output - check timezone settings
+RUN cat /etc/timezone
+
+# clone and checkout desired tesseract version tag
+RUN git clone ${TESSERACT_REPOSITORY} /build_tesseract
+WORKDIR /build_tesseract
+RUN git checkout ${TESSERACT_REF}
+
+# compile
+RUN ./autogen.sh && ./configure --disable-openmp --disable-shared 'CXXFLAGS=-g -O2 -fno-math-errno -Wall -Wextra -Wpedantic'
+RUN LDFLAGS="-L/usr/local/lib" CFLAGS="-I/usr/local/include" make
+RUN make
+RUN make install && ldconfig
+
+# install ocrd-tesserocr (until here commands for installing tesseract-ocr)
 ARG VCS_REF
 ARG BUILD_DATE
 LABEL \
@@ -7,7 +68,6 @@ LABEL \
     org.label-schema.vcs-url="https://github.com/OCR-D/ocrd_tesserocr" \
     org.label-schema.build-date=$BUILD_DATE
 
-ENV DEBIAN_FRONTEND noninteractive
 ENV PYTHONIOENCODING utf8
 
 # avoid HOME/.local/share (hard to predict USER here)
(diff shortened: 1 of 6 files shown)


stweil:
I wonder whether there are still reasons for building the `tesseract` binary.

Using the package from a recent Linux distribution is simpler and would save significant build time.

Another possible approach would also work for `tesserocr` and some more parts of OCR-D: OCR-D could use its own package repositories for all parts with simple dependencies.

bertsky:
> I wonder whether there are still reasons for building the `tesseract` binary.
> 
> Using the package from a recent Linux distribution is simpler and would save significant build time.

Because most of the time, we cannot use Tesseract from a Linux distribution: our base distro is usually older than the current one, and we have no control over Tesseract features that we actually need. The same goes for PPA.

We had good reasons to pin to a specific Tesseract version via source build in subrepo. No reason to give that up now.

> 
> Another possible approach would also work for `tesserocr` and some more parts of OCR-D: OCR-D could use its own package repositories for all parts with simple dependencies.

Much simpler: [conda](https://github.com/OCR-D/core/pull/1072)

joschrew:
@kba: Your changes resolved all my erros with my test workspace. I added a resmgr call to the dockerimage to add eng traineddata. I get an error when trying to process without it.

Dockerfile
```diff
@@ -7,32 +11,78 @@ LABEL \
...
+COPY repo/tesseract ./repo/tesseract
 COPY Makefile .
-RUN make deps-ubuntu && \
-    apt-get install -y --no-install-recommends \
-    g++ \
-    && make deps install \
-    && rm -rf /build-ocrd \
+RUN apt-get install -y --no-install-recommends \
+    python3 \
+    python3-pip \
```
bertsky, line 75: I don't think we need this – as we already use ocrd/core as base img.

Dockerfile
```diff
@@ -7,32 +11,78 @@ LABEL \
...
 COPY setup.py .
 COPY ocrd_tesserocr/ocrd-tool.json .
 COPY README.md .
 COPY requirements.txt .
 COPY requirements_test.txt .
 COPY ocrd_tesserocr ./ocrd_tesserocr
+COPY repo/tesserocr ./repo/tesserocr
+COPY repo/tesseract ./repo/tesseract
 COPY Makefile .
-RUN make deps-ubuntu && \
```
bertsky, line 26: I would really prefer to keep using `make deps-ubuntu`, i.e. encapsulate knowledge of the system dependencies in the Makefile rather than the Dockerfile.

Dockerfile
```diff
@@ -7,32 +11,78 @@ LABEL \
...
+	tzdata \
+	xzgv \
+	zlib1g-dev \
+	libleptonica-dev \
+	libpango1.0-dev \
+	libicu-dev \
+	autotools-dev \
+	automake \
+	libcurl4-nss-dev \
+	libarchive-dev
```
bertsky, line 48: Could you please move these to the `deps-ubuntu` target (and remove the PPA stuff there)?

Dockerfile
```diff
@@ -7,32 +11,78 @@ LABEL \
...
+    && rm -rf /build-ocrd_tesserocr \
     && apt-get -y remove --auto-remove g++ libtesseract-dev make
+
 # PPA tessdata prefix (= ocrd_tesserocr moduledir) is owned by root
-RUN sudo chmod go+w `dpkg-query -L tesseract-ocr-eng | sed -n s,/eng.traineddata,,p`
+# next line causes failure because tesseract-ocr-eng not existing. Not sure if needed, so skipping
+# RUN sudo chmod go+w `dpkg-query -L tesseract-ocr-eng | sed -n s,/eng.traineddata,,p`
 RUN ocrd resmgr download ocrd-tesserocr-recognize Fraktur.traineddata
 RUN ocrd resmgr download ocrd-tesserocr-recognize deu.traineddata
+RUN ocrd resmgr download ocrd-tesserocr-recognize eng.traineddata
```
bertsky, line 85: Yes, we should definitely add equ and osd here as well. The old (PPA-based) procedure, and the ocrd_all (wget based) procedure both installed eng+osd+equ models.

bertsky (changes requested):
Great that this is working now.

Some cosmetic change requests below. Adapting CircleCI config should follow.

bertsky:
> Adapting CircleCI config should follow.

In fact, since it already seems broken on master – unfortunately CircleCI does not keep the logs long enough, but I guess it's about the TESSDATA_PREFIX / resmgr location – we should fix this here.

So I suggest (after rewriting `deps-ubuntu` as proposed above) to update the CircleCI config to do `make install-tesseract install-tesserocr` before `make install`.

Dockerfile
bertsky: thx!

Dockerfile
bertsky: thx!

Dockerfile
bertsky: thx!

Dockerfile
bertsky: thx!

bertsky:
> In fact, since it already seems broken on master – unfortunately CircleCI does not keep the logs long enough, but I guess it's about the TESSDATA_PREFIX / resmgr location – we should fix this here.
> 
> So I suggest (after rewriting `deps-ubuntu` as proposed above) to update the CircleCI config to do `make install-tesseract install-tesserocr` before `make install`.

Now the CI config definitely needs `make install-tesseract install-tesserocr`. Also, we must drop the `chmod` workaround (for which there is no need anymore).

bertsky:
> Now the CI config definitely needs `make install-tesseract install-tesserocr`. Also, we must drop the `chmod` workaround (for which there is no need anymore).

@joschrew do you want me to make that change (on your fork's writable branch)?

bertsky (approved):
At last!

bertsky:
Oh, maybe we should also migrate `make install tesseract-training` here? (Once we remove these rules from ocrd_all, there would be no more way to compile `lstmtraining`, `combine_tessdata` etc.)

PR #199, opened 2024-02-18
convert tests to pytest, make coverage works again

kba:


.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - run: ocrd resmgr download ocrd-tesserocr-recognize deu.traineddata
       - run: ocrd resmgr download ocrd-tesserocr-recognize Fraktur.traineddata
       - run: make test-cli
```
bertsky, line 29: We're still missing `make test`. And it now fails (locally)...

test/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+from ocrd_utils import pushd_popd
+from pytest import fixture
+
+from test.assets import assets as assets
+
+METS_KANT_BINARIZED = assets.url_of('kant_aufklaerung_1784-binarized/data/mets.xml')
+METS_HEROLD_SMALL = assets.url_of('SBB0000F29300010000/data/mets_one_file.xml')
+
+@fixture
+def workspace_kant_binarized():
```
bertsky, line 11: In removing test/base.py, you also removed setting up the logging system. This leaves me with **no** logs now.

bertsky (commented):
I am getting lots of failed
```
FileNotFoundError: File path passed as 'url' to download_to_directory does not exist: 'OCR-D-IMG/OCR-D-IMG_0017.tif
```
now. So apparently, somehow the `local_filename` branch of `workspace.download_file` was not successful. 

But first we need to get logging operational again.

PR #200, opened 2024-02-22
Fix build and tests

bertsky:
follow-up to #197 and #199

~~first step: show that CI currently fails~~

I had to add `download=True` anyway. It now succeeds.


mikegerber:
Fixed #202 for me, for a manual install of Tesseract (`make install-tesseract` from here)

bertsky:
d5724ce was for [the ](https://github.com/sirfz/tesserocr/pull/343) (which has now been merged, so we could even go to tesserocr:master).

kba (approved):

PR #203, opened 2024-03-13
Rename build directory for tesseract

stweil:
Using build/tesseract instead of build_tesseract simpifies things for other projects like ocrd_all and makes a cleaner directory structure.


stweil:
@kba, please merge.

stweil:
Ping?

PR #215, opened 2024-07-02
CircleCI: add CD via PyPI

bertsky:
Not sure if there is a better way. We cannot use our own Docker image in Github Actions, but in CircleCI we can. (Without that image, I am not sure if the build could succeed without first installing Tesseract and tesserocr.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,9 @@ workflows:
             parameters:
               python-version: ['3.7', '3.8', '3.9', '3.10']
   deploy:
+    when:
+      condition:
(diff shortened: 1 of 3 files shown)


bertsky:
So deploy-pypi of course can only work from upstream (due to the credentials), not from forks. The same is true for deploy-docker. Hence the last 3 commits:

- 1e54bd5 ensures that GH Actions pushing to GHCR will only push to the respective repo name (like `ghcr.io/bertsky/ocrd_tesserocr`), so this job can always run on master
- fde2c2e and 067ad52 ensure that CircleCI does not run the `deploy` jobs on forks

PR #216, opened 2024-07-06
Migrate ocrd v3

bertsky:
Depends on OCR-D/core#1240 (i.e. `ocrd>=3.0`).


ocrd_tesserocr/binarize.py
```diff
@@ -6,35 +6,26 @@
...
+    @property
+    def executable(self):
+        return 'ocrd-tesserocr-binarize'
+
+    def _init(self):
+        # use default model (eng) with vanilla tesserocr API
+        self.tessapi = PyTessBaseAPI()
 
-    def process(self):
+    def process_page_pcgts(self, pcgts, output_file_id=None, page_id=None):
```
kba, line 25: Besides missing the typing, this has a different signature, pcgts is not variadic here. Python allows this and it is convenient but I am wondering whether it would be better to be consistent with the typing?

```suggestion
    def process_page_pcgts(self, *input_pcgts, output_file_id=None, page_id=None):
        pcgts = input_pcgts[0]
```

bertsky: Typing – sure, I was just to lazy again.

Variadic – I thought it would be clearer like that. So we would not have to do any arity checking in the function itself – a type checker could simply detect invalid use cases which do pass multiple pages at once. But perhaps I am wrong. (Also, we already have the arity assertion in `setup`.)

kba:
Updated and ready for release I think.

PR #218, opened 2024-07-26
Update dockerfile for core 2.67.1

joschrew:
includes: 
- update version of ocrd-core base image to latest core version
- fix work/build-dir usage and removal. Build dir in core was changed to /build/core
- remove ocrd-network logfiles, which are created through resmgr call

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ocrd/core:v2.62.0 AS base
+FROM ocrd/core:v2.67.1 AS base
 # set proper locales
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
@@ -31,7 +31,7 @@ ENV XDG_DATA_HOME /usr/local/share
 ENV XDG_CONFIG_HOME /usr/local/share/ocrd-resources
 ENV TESSDATA_PREFIX $XDG_DATA_HOME/tessdata
 
-WORKDIR /build
+WORKDIR /build/ocrd_tesserocr
 COPY setup.py .
 COPY ocrd_tesserocr/ocrd-tool.json .
 COPY README.md .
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ocrd/core:v2.62.0 AS base
+FROM ocrd/core:v2.67.1 AS base
```
stweil, line 1:
```suggestion
FROM docker.io/ocrd/core:v2.67.1 AS base
```

stweil: While updating the version, support for podman can be added, too.

bertsky: Until now we always use Dockerhub registry implicitly, so supporting Podman would require similar changes in lots of places. Also, we don't know of Dockerhub will stay free. Therefore I'd prefer a more general mechanism.

How about generating derived Dockerfiles from the existing, git-controlled Dockerfile via makefile rule? For example, we could have:

```make

Dockerfile.podman: Dockerfile
	sed "s,FROM ocrd/core,FROM docker.io/ocrd/core," < $< > $@
	#sed "s,FROM ocrd/core,FROM ghcr.io/ocr-d/core," < $< > $@

podman: Dockerfile.podman
	podman build -f $< .
```

bertsky (approved):
LGTM – thx!

Dockerfile
stweil: The suggested change would not change the status quo for Docker users, but add support for Podman here without adding complexity. ocrd_olena already uses this pattern, 13 other Dockerfiles would still need it.

bertsky: Sure, we can also do that change along the way (and come back to a more general mechanism when really needed).

stweil (approved):
Thanks!

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ocrd/core:v2.62.0 AS base
+FROM docker.io/ocrd/core:v2.67.1 AS base
```
bertsky, line 1:
```suggestion
FROM docker.io/ocrd/core:v2.67.2 AS base
```

PR #219, opened 2024-08-20
Fix broken docker image

joschrew:
The current ocrd/tesserocr-image is broken. When I run `docker run --rm ocrd/tesserocr ocrd-tesserocr-segment-region --help` I get:
```
Traceback (most recent call last):
  File "/usr/local/bin/ocrd-tesserocr-recognize", line 5, in <module>
    from ocrd_tesserocr.cli import ocrd_tesserocr_recognize
ModuleNotFoundError: No module named 'ocrd_tesserocr'
```
I think when modifying the image lately the copy command was changed in the wrong way, which is fixed by this pr

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,19 @@ jobs:
       - run: make coverage
       - codecov/upload
 
+  build-docker:
+    docker:
+      - image: circleci/buildpack-deps:stretch
+    environment:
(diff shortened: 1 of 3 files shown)


bertsky (approved):
Wow. My mistake!

Thanks for spotting, reporting and fixing!

bertsky:
I forgot to include a smoke test for the built Docker images in the CD.

PR #220, opened 2025-04-18
make docker: allow overriding $(DOCKER) command (e.g. podman)

kba:


Makefile
```diff
@@ -102,7 +103,7 @@ deps-test:
 
 # Build docker image
 docker: repo/tesseract repo/tesserocr
-	docker build \
+	$(DOCKER) build \
```
bertsky, line 106: It would require changes in a lot of places to follow through with this.

Wouldn't it be easier to `alias docker=podman` on systems where one wants to replace one with the other?

bertsky (approved):

Makefile
stweil: That would not work. Aliases are not inherited by child processes.

bertsky: > That would not work. Aliases are not inherited by child processes.

...or shell functions defined in .profile, or symlinking into PATH...

kba: I have been using that pattern in our local deployments, to test locally with podman and build/deploy with docker. I could create similar PR while I clean up my local build branches.

>> Aliases are not inherited by child processes.
> ...or shell functions defined in .profile, or symlinking into PATH...

I usually do

```
echo '#!/usr/bin/env bash' > ~/bin/docker
echo 'podman "$@"' > ~/bin/docker
chmod a+x ~/bin/docker
```

but that gets confusing when both docker and podman are installed.

bertsky: sigh, yes, that's a contingency one cannot rule out, I guess. So another round of module updates, it is...

Makefile
```diff
@@ -122,7 +123,7 @@ $(TESSERACT_PREFIX)/bin/tesseract: build/tesseract/Makefile
 $(TESSERACT_PREFIX)/bin/lstmtraining: build/tesseract/Makefile
 	$(MAKE) -C build/tesseract training-install
 
-TESSERACT_CONFIG ?= --disable-openmp --disable-shared CXXFLAGS="-g -O2 -fPIC -fno-math-errno -Wall -Wextra -Wpedantic"
+TESSERACT_CONFIG ?= --disable-openmp --disable-shared CXXFLAGS="-g -O2 -fPIC -fno-math-errno -Wall -Wextra -Wpedantic -UNDEBUG"
```
stweil, line 126: @bertsky, it's a bit strange if you add an unrelated commit to a pull request which was already reviewed by yourself.

I am not sure that your change in line 126 is a good idea (debug code adds instructions that increase processing time), and you have not given a proper explanation of why you made it.

stweil: I'm afraid your change will do exactly what you wanted to avoid.

bertsky: see [this](https://github.com/tesseract-ocr/tesseract/blob/main/src/ccutil/errcode.cpp#L72-L83) why we must at least avoid `NDEBUG`; keeping `-g` otherwise is a good idea to get symbols when there is a crash

all this is a far cry from being satisfactory (libtesseract must use exceptions in the end!), just less insane (segfault)

stweil: So you prefer getting a SIGSEGV because of your change instead of getting an abort() like in the current configuration?

bertsky: in other places like [this](https://github.com/tesseract-ocr/tesseract/blob/68caca6951b384ef5882e98b64198964a37af488/src/ccutil/elst.h#L563-L566) we need `!defined(NDEBUG)` for certain safety checks

bertsky: > So you prefer getting a SIGSEGV because of your change instead of getting an abort() like in the current configuration?

No, segfault is obviously even worse, but I'm in a bind: the place that actually triggers my latest segfault/abort needs `!defined(NDEBUG)` – so the snippet I showed first behaves different, but is (currently) not as relevant

bertsky: See https://github.com/sirfz/tesserocr/pull/365 for how I would like this to be solved.

PR #221, opened 2025-07-15
fixing https://github.com/qurator-spk/eynollah/issues/169 : This fix actually resolves the error. Additionally, the contour with intersections no longer appears in the results.

kba:
For discussion: @vahidrezanezhad has a workaround for the multipolygon issue where eynollah produces polygons with self-intersections.

### DIFF
--- a/ocrd_tesserocr/common.py
+++ b/ocrd_tesserocr/common.py
@@ -3,7 +3,7 @@ from PIL import Image, ImageStat
 
 import numpy as np
 from scipy.sparse.csgraph import minimum_spanning_tree
-from shapely.geometry import Polygon, LineString
+from shapely.geometry import Polygon, LineString, MultiPolygon
 from shapely.ops import unary_union, nearest_points, orient
 from shapely import set_precision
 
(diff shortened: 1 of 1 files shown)


ocrd_tesserocr/common.py
```diff
@@ -254,6 +254,10 @@ def make_join(polygons, scale=20):
         bridgep = LineString(nearest).buffer(max(1, scale/5), resolution=1)
         polygons.append(bridgep)
     jointp = unary_union(polygons)
+    
+    if isinstance(jointp, MultiPolygon):
+        jointp = max(jointp.geoms, key=lambda p: p.area)  # Choose the largest polygon
```
bertsky, line 259: One simple way to share example polygon data is by using Shapely's `wkt` serialization, which can easily be printed, plotted, re-imported etc. So here, e.g. if you do...
```suggestion
        for geom in jointp.geoms:
            print(geom.wkt)
        jointp = max(jointp.geoms, key=lambda p: p.area)  # Choose the largest polygon
```
...and then send me the stdout, that would really help.

bertsky (commented):
I can see how this would help – as a workaround.

But really, this is not the way to fix this. The code above in particular tries to avoid yielding multiple geometries – by finding the best path through all parts and connecting them with little "bridges". So if that failed, for some reason, we must fix that directly.

Could you please provide me with the input polygons that caused this?

PR #222, opened 2026-01-14
add segmentation level paragraph

bertsky:
Tesseract allows retrieving paragraphs, so we should also offer this. In OCR-D, we don't usually have this, so offer
- `flat` → paragraphs **as** normal regions
- `recursive` → paragraphs **inside** block-level regions

In both cases, the ReadingOrder will reflect the recursive structure via ordered subgroups.


kba (approved):
LGTM, well documented and tested, sorry for not reviewing earlier.

I'll test this some more after the dependency-update release I'll do now, a release including functional changes like this will follow soon.

PR #224, opened 2026-05-04
docker: use version from git tag

kba:
* Use versioned releases whenever possible (i.e. after release with version as git tag)
* Use GH Actions for all docker image building and pushing - ghcr.io with GHA and docker.io with CircleCI offers no benefits AFAICT
* Do the PyPI releases in a separate GHA workflow with OpenID Connect instead of token-based auth

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,67 +35,11 @@ jobs:
       - codecov/upload
     resource_class: large
 
-  build-docker:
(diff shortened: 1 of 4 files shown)


kba:
Still open for debate but I do need a versioned docker.io release today, so went ahead and released it. If something breaks, it is only related to CI/CD, not functional.

PR #225, opened 2026-06-05
update base image to OCR-D/core v3.13.1

kba:


kba:
Tests pass, only coverage is still running.

